PR #8, opened 2020-11-03
Modifying Code for multiple Hardware Options

kkloesener:
Hopefully this Codebase can be used to modularize this code for different hardware platforms and options.

Please Review!


biologist79:
Looks good but have some doubts regarding SPI for Audiokit. Thing is the default-lib relies on having SPI @default-pins. But as this isn't the case, I've been using a custom lib. It's provided in Audiokit's subfolder, but yes, this hasn't been well documented by myself. In fact I wasn't sure if anybody else would be interested in running my Tonuino-fork using Audiokit. That's why I didn't directly integrate it the way you did. However, it's (at least in my opinion) a tradeoff, as having a growing number of things configurable in the code, makes it hard for "newbies" to understand.

However, will accept but need some time to test. Two questions:
a) Have you been testing running it this way with i2c?
b) How can one decide which platform (lolin32 or a1s) to use? I saw to declared both with platformio.ini but were does the selection take place?

kkloesener:
oh yeah, found still some bugs in my code, i will add additional pull-request for that.

regarding your questions:
a) i'm in progress of testing it completly
b) you can compile and upload for the needed platform within VS-Code with platformIO Extension - see the attached Screenshot
2020-11-04_09h16_33
This way we can even simpler extend the code for multiple Hardware-Choices

biologist79:
Hi,

a) fine
b) Thanks for clarification. Wasn't aware of that feature.

The "hack" with the modified lib is a bit dirty as you cannot simply distinguish between original and modified lib. The are almost the same. Would make much sense to modify the original lib in order to enable to provide a SPI-object. That way it wouldn't be necessary to have both. But I didn't put any efforts in that point so far as I had enough work with Tonuino itself :-)

biologist79:
Don't worry, still have this pull request in my mind :-)

kkloesener:
After much rework with the libraries not able to use TwoWire i ended up getting all the things done!
I hopefully merged everything correct from your master in my working tree.

Please Review

biologist79:
Hi,

thanks for your contribution - makes things a lot easier for me.
Yesterday I merged your code into my devel-branch. Think it's a good decision to proceed with develboard-specific files. Added another ohne for Lolin D32 and guess there will follow some (not yet checked in). However, RFID didn't work with 2 SPI-instances. Don't know why, RC522's lib didn't complain. Will need to have a look again this evening.

biologist79:
Decided to commit my stuff instantly to my devel-branch: https://github.com/biologist79/Tonuino-ESP32-I2S/tree/devel
However, RFID is not yet working. Hate this lib.

biologist79:
I just comitted some changes.
a) RFID with RC522 and two SPI-instances is working now
b) I dissolved the nested structure of preprocessor-directives a bit in setup() as I think there have been dependencies that where not necessary.

Please have a look at my devel-branch and test whether your software still works. Hopefully I didn't change anything that's related to your stuff.

biologist79:
I was told PNS5180 + SD-MMC is working as well.

PR #11, opened 2020-12-02
Feature/filebrowser

mariolukas:
**Changes:**
- added a File Browser to the HTML/JavaScript code
- added functions for creating a file list in main.cpp
- added websocket endpoints for file list handling
- added a static route for delivering the files.json file to the frontend File Browser
- refactored user notifications in HTML/JavaScript by introducing  toastr

**How it works:**
Just open the HTML page your Browser and click the refresh icon below the file list. 
The list will be generated. Once generated there is no need for a recreation until the 
state of directories/files changed (e.g.  new files or folders  are added, renamed or deleted).

The file list appears automatically after it was created. Every page reload will automatically
load the files.json list for creating the file tree in the file browser. 

**Additional notes:**
Some files like platformio.ini might contain some specific changes to my ESP32 dev Board
i  am using  [az-delivery-devkit-v4](https://www.az-delivery.de/products/esp32-developmentboard) Boards.
Also some pin mappings are customized to fi my Tonuinos. (Rotary Encoder pins are switched and number of Leds are changed).

**Todo:**
- hide file browser when no file list was found and show a bigger refresh file list creation link instead of the refresh icon
- alternative approach:  create an  empty file automatically in setup() when no files.json was found


mariolukas:
Something went wrong with my fork. Seems that the whole files are marked as replacement, instead of marking  only the changes. I can try to start by a new fork and include all the changes,  otherwise you have to cherry pick all the changes i made. Starting with a new fork will take some time.

PR #26, opened 2020-12-13
Add support for SD MMC 1 Bit mode & PN5180 reader

tueddy:
SD MMC 1 Bit mode:
This mode doubles the SD card speed (FTP upload speed ~350 kB/s compared to ~180 kB/s in SPI mode). Needs one PIN less but needs fixed PIN assignments

Add support for the PN5180 RFID reader:
PN5180 has better RFID range/sensitivity and can read ISO-15693 / iCode SLIX2 aka Tonies


tueddy:
Seems to work, close now..

PR #34, opened 2020-12-19
Abstraction of SD and SD_MMC with FS

grch87:
Thanks to tueddy the 1Bit SD_MMC Mode was introduced, which allows me to use the main.cpp from the master without modification. This makes my life easier :)
In order to make the maintenance of the this feature easier, I introduced an abstraction of SD and SD_MMC via FS. This reduces the compiler switches down to the include and initialisation.

I was only to test it with SD_MMC here, but there it seems to work (FTP, Audio, Webserver).


biologist79:
Merged ccf4c78 into my master-branch.
Thanks for your contribution!

PR #39, opened 2020-12-21
Minor Improvements

grch87:
Hi,
just gathered some small changes. Containing a bug fixes, an improvements and a feature
Starting with the bug fix:
The return code of the connecttofs function is now evaluated. In case of an error the track is considered as finished and led error is shown. I had before the problem that files with 'Umlaute' causes problems in the audio lib (created an issue for that) and the files can't be opened. The tonuino was then frozen.

The improvement tackles the voltage measurement. Now several analog reads are taken and the average value is used for the voltage calculation. So the noise is reduced. Currently 20 values are used, as it is accurate enough I think (+-0.05V). Note: One analog read takes 10µs, if you want to have more measurements. I saw that tipp in a youtube video of Andreas Spieß (Guy with the swiss accent).

The feature is that, when you press the previous track button and the current track is already older than 2s the current track is played instead of the previous track. This is then similar to other audio playing devices. But I'm not sad, if this is not making it into the master tree :)

Furthermore I updated the platformio.ini and uses in the build configs so called dynamic varibales, primarily for the lib deps. As the lib deps weren't visible for the D32 Pro build config.


biologist79:
Had a first look at it: really cool! 
I'm fine with this 2s-feature but maybe I'd reduce it down to 1s. We'll see.

biologist79:
Cherry-picked 2efa10a and 5e753ba

PR #45, opened 2021-01-05
Add support for PN532 RFID-Reader

datenzar:
Since I prefer the PN532 RFID-Reader (more versatile, can even be used with UART), I enhanced the code to support that chip.

Feel free to integrate and thanks a lot for this project!

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -21,6 +21,7 @@ lib_deps_external =
 	ESP Async WebServer
 	https://github.com/me-no-dev/AsyncTCP
 	https://github.com/bblanchon/ArduinoJson.git
+	https://github.com/The-MEO/PN532.git#platformio	# library for PN532
 ; https://github.com/pschatzmann/ESP32-A2DP.git
 
 [env:common]
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -50,9 +50,8 @@
     #include <SPI.h>
     #include <PN532_SPI.h>
     #include "PN532.h"
-    
-    PN532_SPI pn532spi(SPI, 5);
-    PN532 nfc(pn532spi);
+    static PN532_SPI pn532spi(SPI, RFID_CS);
+    static PN532 nfc(pn532spi);
 #endif
 #ifdef RFID_READER_TYPE_PN532_I2C
     #include <Wire.h>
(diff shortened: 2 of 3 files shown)


biologist79:
Hi,
thanks for your contribution. 
Need to get such a rfid-reader at first in order to test it. At least I'll need to change GPIO5 to a variable. 

But: what exactly is PN532 doing better in comparison to RC522? Have been testing PN5180 and it's advantage is better sensitivity (but needs some more GPIOs).

datenzar:
Hi,

good point about the change of GPIO5. I used it instead of 21 since the pins all fit in one DuPont shell on the LOLIN32 but of course, it should be subject to configuration.

As far as the comparison to PN5180 is concerned, I cannot tell much. Compared to the standard RC522 it's the size that was most important to me (using [this one](https://www.amazon.de/ICQUANZX-Nahfeldkommunikationsleser-Modul-Kit-Schl%C3%BCsselkarte-DIY-Smartphone-Android-Handy/dp/B07VT431QZ/ref=sr_1_3?__mk_de_DE=%C3%85M%C3%85%C5%BD%C3%95%C3%91&dchild=1&keywords=PN532+NFC&qid=1609846584&sr=8-3).

Apart from that, you can also connect it via UART (GPIO21/GPIO22) and hence save the SPI Pins for different uses.
I quickly measured the sensitivity and it starts to recognize the RFID tags with 3cm if nothing is in between.

tueddy:
Hi,
i have all 3 readers tested, MFRC522, PN532 and PN5180.
MFRC522 and PN532 are very similar in RFID range and sensibility. There is no big difference.

The PN5180 brings  a huge step forward but as said before at cost of 2 more pins (and 3 if you want to wakeup from deep sleep with low power card detection LPCD).
The PN5180 can detect ISO-14443 cards >5 cm and ISO-15693 > 8cm.

datenzar:
Thanks for your feedback @tueddy! I already ordered a PN5180 board from Aliexpress 2 weeks ago and it's on its way from China.

However, regarding the form factor, I couldn't find anything as small as that PN532 board. Maybe someone can give a hint.

datenzar:
Hmm,  that's indeed weird. I don't work on a private repository. You can check the branch at https://github.com/datenzar/Tonuino-ESP32-I2S/tree/pn532

biologist79:
Strange. Anyway, now I know where to get it from :-)

biologist79:
I just ordered my pn532 a few days ago in .cn. Will take it's time to arrive... :-)

biologist79:
My pn532 has finally arrived today. Will test within the next days...

datenzar:
Very nice, Today I actually received a package of 2 as well. I guess they shared the same shipping container ;-)

PR #72, opened 2021-01-26
File browser replacement and delete fix

grch87:
I improved the deletion and moved the recursive folder deletion logic to the server. The solution to do that on the client side was crap.
I moved the new filebrowser to the rfid tab as requested and replace Marios current filebrowser.


biologist79:
Already integrated by cherry-picking.

PR #87, opened 2021-02-14
reorg RFID page. nicer butons

ceear:


biologist79:
Added via cherry-picking.
Thanks a lot for your contribution!

PR #91, opened 2021-02-16
reintroduce arrow and placeholder in the dropdown box. switch to music tab when browsing files. add percent to progressbar.

ceear:
reintroduce arrow and placeholder in the dropdown box. switch to music tab when browsing files. add percent to progressbar.


biologist79:
Perfect! Thank you for your contribution!

PR #99, opened 2021-03-19
Add automated build and status badge

tuniii:


tuniii:
This PR adds an automated build via Github Actions for all the supported boards. A status badge for the build workflow has been added to the README at the root. This badge shall help to keep track of the overall build status.

tuniii:
The badge will look like this: 
image

biologist79 (commented):
readme + platformio.ini is ok.
settings.h doesn't make sense for me.
a) HAL is always set.
b) You defined, that if i2c isn't set für RC522, SPI will be used instead. But what if an user wants to use PN5180?

biologist79:
Merged via cherry-picking.
Thanks for your support!

tuniii:
I mainly made the changes in settings.h to make the board esp32-a1s build again, because the build is currently broken.
a) if the HAL is set via the build_flags it is then overwritten again with settings.h. Therefore I added the #ifndef.
b) The current settings.h is conflicting for the board esp32-a1s and the RFID_READER_TYPE_MFRC522_I2C set via the build_flags. A #undef at the end of the settings.h could also help here.

biologist79:
Regarding the other stuff: build-process doesn't work.
But unfortunately, github isn't very verbose. It just says "Error:" for ALL tasks.

tuniii:
esp32-a1s build  is failing. See my previous comment.
image

biologist79:
I'll remove that flag.

biologist79:
Regarding a) Ok. Wasn't aware of that. Will add it.
But b) is to addressed somehow different. Will remove that i2c-flag from platformio.ini. But in this case I need GPIOs for RFID-SPI.

tuniii:
Somehow it is mandatory somewhere, because only the MFRC522_I2C_Library lib is included at the board definition.
I see two options:
1. ) Set the preprocessor #define for I2C at platformio.ini and #undef the other rfid readers (e.g. at the board specific settings file)
2. ) #undef all readers at the board specific settings file and #define the I2C hardcoded after the #undef

tuniii:
Ok. I see. You have chosen option three 😂.

biologist79:
Hrhr. Guess it's not easy to make it bullet-proof.
Back in the days I had it running with:
SD (SPI)
RFID (SPI)
Both shared the same SPI-instance.

However, this was before ESPuino learned to handle SD_MMC. With having SD_MMC in place, it's not possible to share SPI anymore. But A1S doesn't provide enough GPIOs for
SD_MMC
RFID-SPI
Rotary encoder
3 Buttons (unless analogRead() is used).

Is that correct?
To be honest: I own this board but don't have time to test it completely :-(

biologist79:
However: still don't know why the build-process doesn't run :-(

tuniii:
I can also not tell, because I do not own the board. I made just some suggestions from the source code. You need to add the RFID lib I guess. Currently only the I2C RFID lib as added.

PR #101, opened 2021-03-25
rework actual codebase for rfid via i2c

kkloesener:
some smaller changes needed to get i2c support correctly compiled


biologist79:
Thanks for providing that fixes!
But got a question: what is numbers 198 + 199 for buttons? I know it's not new and in the past I didn't really care. But now I'd like to know :-)
For ESP32 theres 0->39 (GPIOs) that were recently extended by 100 -> 115 (port-expander). 99 should be used to disable.

kkloesener:
They are just some placeholders at this time.

PR #103, opened 2021-04-16
unify defines and eng spelling

phschoen:


biologist79:
Thanks for your contribution.
In general, this is ok. But currently a code-refactoring is in progress. That said: makes sense to wait until it's complete. But especially main.cpp will be different. That different, that this patch can't be applied.

However, I'll leave it open in order to not forget it :-)

phschoen:
ok ;) but speaking of refactoring i did not find any branch for it or is it only local for now?
I would also to prefer to move function to "modules" with separated hpp/cpp files to get a better structure/interface separation compared to just one big main.cpp. 

I also started to work within the rfid section to generate gone event in case a card is not reachable any more to have the feature to stop if the card is a playlist card as an option. As well I added a MPU6050 in order to get a tilt and taping event for additional inputs and i am soon finished with them but have to clean it up. Should I wait with this also after the refactoring?

biologist79:
It's currently local only as it's under development. Will be modules. It's described here: https://forum.espuino.de/t/rfid-mit-oder-ohne-task/353/8
So better wait until the new structure is released + all testings are finished. As I currently don't have enough time to do the refactoring @tuniii will do that. And while this task is in progress I promised to not check-in new code :-)

I never expected this project to become that big. However, makes sense to split it into modules.

tuniii:
Refactoring is in progress: [https://github.com/tuniii/ESPuino/tree/refactoring](https://github.com/tuniii/ESPuino/tree/refactoring)

Most of the functions have been moved to separate modules. But work is still in progress.

biologist79:
Added all the stuff manually to refactoring-branch.
Thanks for your input.

Hint: will not be added to the current master-branch as this one is going to be deprecated soon.

PR #104, opened 2021-04-17
Add support for PN532 RFID-Reader

datenzar:
Sorry but I closed original PR #45

Here's now the updated version. The PN532 is now working in the box I completed for my daughter.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -25,6 +25,7 @@ lib_deps_external =
 	https://github.com/Arduino-IRremote/Arduino-IRremote.git
 	https://github.com/kkloesener/MFRC522_I2C.git
 	https://github.com/miguelbalboa/rfid.git
+	https://github.com/The-MEO/PN532.git#platformio	# library for PN532
 
 [env:common]
 platform = espressif32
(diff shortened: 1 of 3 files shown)


kkloesener:
Please review your Code for I2C Communication
There are already TwoWire Instances for different use-cases defined. This could conflict when using the default Wire-Object.
Please also have a look at the refactoring branch. I think new Pull Requests will only get merged with the new structure in the near future.

biologist79:
Please consider the main-branch as "dead". Unfortunately I don't have time to maintain both. Old implementation is still there as refactoring-branch is currently being tested.
@datenzar I'm sorry, I know you've already been waiting long to get pn532 integrated. Thing is that (at least from my perspective) there are lots of points that are more important

biologist79:
Meanwhile I removed tasks from PN5180 and RC522 as well in refactoring-branch in order to get rid of overhead. In case of PN5180 this led to problems because it took to long to scan for rfid. @tunii addressed this by adding a state-machine.

Think we need to investigate this for PN532 as well.

biologist79:
Finally decided to not support PN532 because it has no real benefit over PN5180 and/or RC522.
RC522 is also possible with i2c (special rfid-board is necessary).

PR #105, opened 2021-04-17
WIP: Add support for touch sensors

datenzar:
Since the ESP32 supports capacitive GPIOs, I replaced the NEXT & PREV buttons with the capacitive sensors. Both GPIOs are wired to the bolts of the case. I will provide a photo shortly

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -25,6 +25,7 @@ lib_deps_external =
 	https://github.com/Arduino-IRremote/Arduino-IRremote.git
 	https://github.com/kkloesener/MFRC522_I2C.git
 	https://github.com/miguelbalboa/rfid.git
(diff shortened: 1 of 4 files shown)


datenzar:
Please note: This PR is still not complete. I have a problem, where the recognized touch will allways skip right to the end of the playlist (instead of moving to the next track). I'm still troubleshooting but any hint would be appreciated...

biologist79:
Hi,

thanks for your contribution. Problem is that a code-refactoring is in progress (smaller main.cpp with modules). So your commit will have to be refactored again when it's done.
Preview: https://github.com/tuniii/ESPuino/tree/refactoring

biologist79:
Closed this as I considered it being stale + don't have time to support it continously.

PR #116, opened 2021-11-08
Fix PN5180 sleep issues

SZenglein:
Button wakeup will now be registered in case of spurious wakeups
Correctly wake up on LOW signal


SZenglein:
See https://forum.espuino.de/t/pn5180-deepsleep/688

@tueddy please test and confirm it fixes wakeup issues

biologist79 (approved):

PR #122, opened 2021-11-21
Add example for 8MB flash devices

SZenglein:


biologist79:
Have not possibility to test but looks good.
Accepted!

PR #132, opened 2021-12-12
Port_Init() must be called to enable amplifiers even when no port extpander is present.

herm:
On my board the amplifier was never enabled because neither `PORT_EXPANDER_ENABLE` nor `HEADPHONE_ADJUST_ENABLE` was defined.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -214,9 +214,7 @@ void setup() {
         Serial.println(F("UNKNOWN"));
     }
 
-    #ifdef PORT_EXPANDER_ENABLE
-        Port_Init();
-    #endif
+    Port_Init();
     Ftp_Init();
     AudioPlayer_Init();
     Mqtt_Init();


biologist79:
Thanks for for contribution. Which board did you use?
However, I'd like to solve it in a different manner. Please revert your change and try adding at https://github.com/biologist79/ESPuino/blob/master/src/AudioPlayer.cpp#L165

`
#ifndef PORT_EXPANDER_ENABLE
#ifdef GPIO_PA_EN
   while (millis() < 500) {
      delay(10);  // Wait a bit in order to avoid ugly noises when enabling amp
   }
   Port_Write(GPIO_PA_EN, true, true);
   #endif
#endif
`

This won't be the final fix but the core of it. Would like restructre the function a little bit in order to avoid code-duplication.
So please give me feedback if this would basically fix your issue.

herm:
This fixes the issue as well. 
However the current `Port_Init` handling is quite strange. The only call to Port_Init is wrapped in an `#ifdef PORT_EXPANDER_ENABLE` and inside this function is another `#ifdef PORT_EXPANDER_ENABLE` but only for small parts of the function. Also currently the behavior of the amplifier enables is different depending on if a port expander is enabled. It doesn't matter if the amplifier enables are connected to the expander or directly (or even if anything is connected to the port expander!).

About my board: It is a custom design based on a chinese ESP32 board with battery charger, an MAX98357A and an MFRC522. It's basically only a prototype based on things I had lying around. I started a similar project like yours, but sometime during development I found your project in the ESP-audioI2S bug tracker and switched because your project is more advanced. (I probably didn't find it during my initial search because based on the name I expected an Arduino board with an ESP and skipped looking at it.)
My hardware is a bit different from the standard design for this project. I'm not switching off the amplifier or card reader using a MOSFET, but use the built-in low power modes instead. One thing my code was able to do which this code isn't yet was low power card detection with the MFRC522. I'm waking up once a second, check if a card is present (in about 3ms) and go to deep sleep again. For maximum power savings you don't want to switch everything else on together with the MFRC522 because on SD cards draw a lot of power during initialisation. So I used the MFRC522 soft power down mode which consumes only 10µA. I also only wake up the MAX98357A while audio is playing and put it back to sleep (0.6µA) afterwards or while audio is paused. That's the reason I needed the amplifier enable to work (even if is currently powers the amplifier on much longer than required). I want to port these changes over to this project once I find the time to do so.

PR #135, opened 2022-01-05
add web GUI option to start FTP server

### LINKED ISSUE #134: Enable FTP via Web UI
Thanks a lot for this project! When I first heard about Tonuino I set out to built my own version that supports proper file handling and also remote upload capability. Luckily I forgot about it for a while and then found ESPuino <3

Anyway, I was wondering if you'd be willing to support FTP Enabling via the Web UI.
There already is a FTP section, so it would only be one more button the UI on a relatively empty section.

Benefit: I don't need to walk to the device and push the button, I can do it all from my PC directly.
I would try to PR if you think this is an idea you like

anoff:
fixes #134
image

verified locally by clicking the button, checking the logs and opening FTP session :)

Only output in logs seems to be the heap report

```text
Aktuelle Batteriespannung: 3.90 V
no cover image for SD-card audio
Freier Heap-Speicher vor FTP-Instanzierung: 142884
Freier Heap-Speicher nach FTP-Instanzierung: 125508
RSSI: -68 dBm
```

Could not verify on LANGUAGE=EN because build fails (even without these changes)

```text
src/LogMessages_EN.cpp:189:40: error: expected initializer before 'PROGMEN'
     const char unableToTellIpAddress[] PROGMEN = "IP-address can't be announced as there's no WiFi-connection available.";
```

### DIFF
--- a/html/management_DE.html
+++ b/html/management_DE.html
@@ -382,6 +382,16 @@
                 </div>
             </form>
         </div>
+        <div class="container" id="ftpStart">
+            <legend>FTP Server starten</legend>
+            <div>
+                Aktiviert den FTP Server bis zum Neustart des Geräts. (FTP muss aktiviert sein)
+            </div>
+            <div class="text-center">
+                <button type="button" class="btn btn-primary" onclick="ftpStart()">Server starten</button>
+            </div>
+        </div>
+    </div>
     </div>
 
     <div class="tab-pane fade" id="nav-general" role="tabpanel" aria-labelledby="nav-general-tab">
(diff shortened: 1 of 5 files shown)


anoff:
Should I mention this in the FTP section in README as well?

biologist79:
Thanks for your contribution - looks good :-)
Will add it asap.
Regarding your question: no, I'll do that myself. Had already plans to revise it.

anoff:
awesome.

btw, the `processHtml.py` took me a while to figure out. In case you're rewriting towards contributor/developer docs a short notice how the HTML files are dealt with might help the next guy :)

biologist79:
Just checked it yesterday. Technically it's fine but I had to apply some tweaks to the html-stuff.
Just as hint as you mentioned via GUI that FTP has to be enabled: its not necessary to indicate this as FTP-menupoint isn't shown if ESPuino was compiled without FTP-support.

However, thanks a lot for your contribution.

anoff:
Ah smart, didn't think that far..

PR #136, opened 2022-01-05
PN5180: better card recognition for ISO-14443 cards

tueddy:
- remove state RFID_PN5180_NFC14443_STATE_SETUPRF: This is already handled in PN5180 library
- wake-up from deep-sleep also with rotating the encoder (not only push the encoder button)
- fix a bug for serving the cover image (wrong imagesize)
- shorter show ESP-IDF version in console-log

### DIFF
--- a/src/Button.cpp
+++ b/src/Button.cpp
@@ -59,6 +59,8 @@ static void Button_DoButtonActions(void);
 void Button_Init() {
     #if (WAKEUP_BUTTON <= 39)
         esp_sleep_enable_ext0_wakeup((gpio_num_t)WAKEUP_BUTTON, 0);
+        esp_sleep_enable_ext0_wakeup((gpio_num_t)ROTARYENCODER_CLK, 0);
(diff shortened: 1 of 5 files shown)


src/Button.cpp
```diff
@@ -59,6 +59,8 @@ static void Button_DoButtonActions(void);
 void Button_Init() {
     #if (WAKEUP_BUTTON <= 39)
         esp_sleep_enable_ext0_wakeup((gpio_num_t)WAKEUP_BUTTON, 0);
+        esp_sleep_enable_ext0_wakeup((gpio_num_t)ROTARYENCODER_CLK, 0);
```
kkloesener, line 62: you need to check #ifdef USEROTARY_ENABLE here for Line 62 and 63

tueddy: Thank's for reviewing!
OK, this change was wrong, reverted.

biologist79:
Merged.

PR #137, opened 2022-01-23
speedup HTTP file-upload

tueddy:
- avoid writing small chunks to sd-card, instead wait for buffer filled > 4K
- increase buffer to 8K
- avoid delay feeding the watchdog time
- show number of chunks written after upload finished


biologist79:
Merged.

PR #140, opened 2022-02-09
Add power-down method in shutdown/sleep procedure

Splash005:
I measured around **100mA** power draw in deep-sleep with my multimeter on the current branch. 
After adding the following line in the shutdown procedure/call I just measured **30mA** in deep-sleep.

My deep-sleep power draw is still quite high, but that seems to be a standard Dev-Board-Problem.

There seems to be no problem after wake-up and reading the cards.

**My ESPuino-Setup:**
- AZ-Delivery DevBoard (powered by Vin-PIN)
- SD-MMC-Module and Interface
- MAX98357A amplifier
- MFRC522 NFC-Reader

### DIFF
--- a/src/RfidMfrc522.cpp
+++ b/src/RfidMfrc522.cpp
@@ -176,6 +176,7 @@
     }
 
     void Rfid_Exit(void) {
+	     mfrc522.PCD_SoftPowerDown();
     }
 
     void Rfid_WakeupCheck(void) {


biologist79:
Thanks for your PR. Will test and integrate it.
However, never had such a huge draw @deepsleep - even with that AZD-stuff. If I was you I'd consider using "external" unpowering by using mosfet(s).

Splash005:
Did another test:
When in deepsleep, unplugged the amplifier and SD-modul --> no change
When unplugged the nfc-reader it dropped from 30mA to 9-10mA (which seems to be normal, because of Voltage regulator and USB-Controller on the dev-board).

Like the method says, it seems like it is just a light sleep with a current draw of around 20mA for the rfid reader only.
As this is just my prototype-box it will be ok like this.

Thank you for your hints and support!

biologist79:

PR #141, opened 2022-02-14
feature: download file from web-ui

tueddy:
- allow download file from web-ui (context menu)
- speedup loading filelist (large number of files/directories)
- localize "loading.."
- Fix memory leaks (missing file.close() )


biologist79:
Merged!
Thanks for your contribution.

PR #144, opened 2022-02-20
Generic Battery Monitoring

SZenglein:
I cleaned up my previous changes and isolated only the code that paves the way for other battery monitoring solutions (e.g. MAX17055 :grin: ). 
The commit is pretty minimal and I plan to follow it up with an even smaller commit for MAX17055 support. Basically these changes should be unnoticeable in the compiled version. 

**No important functionality changed (no new functionality, led indicators the same, CI should still be happy)**

* Update Battery.h with methods for logging/mqtt and percentage estimation
* Move Battery.cpp to BatteryMeasureVoltage.cpp
* Add Battery.cpp for code that is the same regardless of backend
* Use generic methods for battery in e.g. Led.cpp and Cmd.cpp (code is much cleaner now IMHO)


biologist79:
Merged.
Thanks for your contribution!

PR #145, opened 2022-02-28
Support for Max17055 battery fuel gauge IC

SZenglein:
* Add new flag for I2C that is set depending on other modules
* Add Battery.h implementation for Max17055 IC


biologist79:
Merged. That one for MAX17055 is untested by myself. I'll forward any questions regarding MAX17055 to you :->
Thanks for your contribution!

PR #147, opened 2022-03-13
bugfix for battery check interval

SZenglein:
Fixes the changed template name that I forgot to change in the html files.


biologist79:
Merged.

PR #150, opened 2022-05-22
Battery Indicator for numLedsToLight == 0

SZenglein:
Change the behaviour for nearly empty battery: instead of no change in LEDs (because FastLED.show is never called), blank once.
An alternative would be always showing at least one LED.


biologist79:
Merged!

PR #154, opened 2022-07-22
BUG: Volume initially not set correctly

kkloesener:
Volume is correctly read from NVS for initialVolume or previousVolume but was never correctly set when starting the AudioPlayer_Task. This fixes it.

Tested with and without USE_LAST_VOLUME_AFTER_REBOOT defined

### DIFF
--- a/src/AudioPlayer.cpp
+++ b/src/AudioPlayer.cpp
@@ -298,7 +298,7 @@ void AudioPlayer_Task(void *parameter) {
 
     uint8_t settleCount = 0;
     audio->setPinout(I2S_BCLK, I2S_LRC, I2S_DOUT);
-    audio->setVolume(AudioPlayer_GetInitVolume());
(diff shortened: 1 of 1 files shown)


biologist79:
Cannot reproduce this issue. Already works with and without USE_LAST_VOLUME_AFTER_REBOOT.
Which steps are necessary to reproduce it?

kkloesener:
﻿
﻿It’s necessary to shutdown the esp by resetting, flashing or remove powersource.
In this case the audioplayer would always use InitVolume stored in NVS and would ignore a stored value for LastVolume.

I’m not truly sure but this could also be a timing issue when using platforms like A1S.
So your platform could initialize Audioplayer faster and the point when LastVolume is restored is some time later.
This fix just makes sure that Audioplayer initialization uses the actually set volume instead of always using initVolume from NVS.

biologist79:
Works for me.

PR #161, opened 2022-10-10
revision: Add Git revision to logs

fetzerch:
This change adds a pre build script that generates a `gitrevision.h` header with the current git revision. That is then logged in the serial console and web log.

When comparing log files (for instance to hunt issues) it's useful to know exactly the git version the software has been compiled with. If local modifications have been done, '-dirty' is added to the version so that uncommitted changes can easily be identified.

I actually have / had a few problems when flashing. I couldn't get it to work over USB and also via OTA I had sometimes failures. The git revision helps to clearly see if I'm running the expected version.

We could take this one step further and move the specification of the software revision (`20221005-1`) into the `platformio.ini` and then fully generate a single `revision.h` file with with both defines.

### DIFF
--- a/gitVersion.py
+++ b/gitVersion.py
@@ -1,0 +1,50 @@
+# -*- coding: utf-8 -*-
+#!/usr/bin/env python3
+
+"""PlatformIO script to generate a gitversion.h file and adds its path to
+   CPPPATH.
+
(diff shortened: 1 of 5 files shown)


biologist79:
Added by cherry-picking.
Thanks for your contribution!

PR #164, opened 2022-10-16
web: Fix loading covers with UTF16 encoding

fetzerch:
I noticed that only half of my cover images loaded and debugged the issue down to a problem with the ID3 text encoding for the image description. Even if the description is an empty string, if the text encoding is set to UTF16, an extra 0 has to be skipped when skipping over the image description.

Now all images load fine.

See: https://id3.org/id3v2.4.0-frames, chapter 4.14:

```
 <Header for 'Attached picture', ID: "APIC">
 Text encoding      $xx
 MIME type          <text string> $00
 Picture type       $xx
 Description        <text string according to encoding> $00 (00)
 Picture data       <binary data>
```

And: https://id3.org/id3v2.4.0-structure, chapter 4:

```
Frames that allow different types of text encoding contains a text encoding description byte. Possible encodings:

  $00   ISO-8859-1 [ISO-8859-1]. Terminated with $00.
  $01   UTF-16 [UTF-16] encoded Unicode [UNICODE] with BOM. All
        strings in the same frame SHALL have the same byteorder.
        Terminated with $00 00.
  $02   UTF-16BE [UTF-16] encoded Unicode [UNICODE] without BOM.
        Terminated with $00 00.
  $03   UTF-8 [UTF-8] encoded Unicode [UNICODE]. Terminated with $00.
```

### DIFF
--- a/src/Web.cpp
+++ b/src/Web.cpp
@@ -1332,14 +1332,16 @@ static void handleCoverImageRequest(AsyncWebServerRequest *request) {
     Serial.println(coverFileName);
 
     File coverFile = gFSystem.open(coverFileName, FILE_READ);
-    // seek to start position, skip 1 byte encoding
-    coverFile.seek(gPlayProperties.coverFilePos + 1);
+    // seek to start position
+    coverFile.seek(gPlayProperties.coverFilePos);
+    uint8_t encoding = coverFile.read();
     // mime-type (null terminated)
     char mimeType[255];
     for (uint8_t i = 0u; i < 255; i++) {
         mimeType[i] = coverFile.read();
-        if (uint8_t(mimeType[i]) == 0)
+        if (uint8_t(mimeType[i]) == 0) {
             break;
+        }
     }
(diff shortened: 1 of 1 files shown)


tueddy:
Didn't tested because i have no UTF-16 encoded mp3's but modifications looks fine here.
Can you provide an example mp3? What MP3 generator was used, e.g iTunes?

fetzerch:
Sure, here you go: [Test2.zip](https://github.com/biologist79/ESPuino/files/9795776/Test2.zip)

grafik

I've written a [script to convert mp3s](https://github.com/fetzerch/sync_music/) some years ago (which I btw also find quite useful for ESPuino) and apparently UTF-16 is the default in the underlying id3 library (mutagen): https://mutagen.readthedocs.io/en/latest/api/id3_frames.html#mutagen.id3.APIC.
But also [mp3tag](https://www.mp3tag.de/) has a setting for storing id3 in UTF-16, so it's probably not too uncommon.

biologist79:
Added by cherry-picking.
Thanks for your contribution!

PR #160, opened 2022-10-10
MQTT improvements

fetzerch:
As mentioned in the [forum](https://forum.espuino.de/t/espuno-und-nymea-homeautomation/1354) I worked on a few improvements for MQTT. See the individual commit messages for details.

Since this is my first contribution, please let me know if you want me to change anything (or to split the PR into individual changes).

PR #165, opened 2022-10-17
audioplayer: Generalize handling of title

fetzerch:
This change generalizes the generation of the title that is shown in the Web UI and published over MQTT with the following goal in mind:

- Save CPU cycles and memory since the formatting is just done once.
- The exact same format is used for the Web UI and MQTT.
- The initial title sent over MQTT is now correct and no longer "---".
- Make the code easier to understand and maintain.

I've included the mqtt coverChangedState commit from #160, to save some rebasing hazzle -> Therefore please see the individual commits for a proper diff.


biologist79:
Added by cherry-picking.
Thanks for your contribution!

PR #166, opened 2022-10-18
processHtml.py: Modernize and fix execution

fetzerch:
The processHtml script was silently ignored because `extra_scripts` are launched through Platform IO's SCons build system and this doesn't define __file__. Just dropping the `if __name__` however also didn't work and just causes a cryptic SCons error message.

While at it, this modernizes the script and removes the broken and unused html minify logic.


biologist79:
Added by cherry-picking.
Thanks for your contribution!

PR #167, opened 2022-10-18
processHtml.py: Remove generated files

fetzerch:
Now that generating the HTML*_<LANG>.h files works reliably under Linux, Windows and macOS, there's no need to keep the generated files in the repository.


biologist79:
Added by cherry-picking.
Thanks for your contribution!

PR #169, opened 2022-11-03
MQTT Improvements

fetzerch:
Another bunch of MQTT improvements. See individual commit messages.

Most noticeable is the second commit that fixes a race condition that caused broken MQTT packages and ended in a disconnect. I'm not sure what changed, but I got this almost every time now when I started a Webradio stream.

The third commit fixes the issue we discussed in the forum that `topicTrackState` was sent once too often.

### DIFF
--- a/src/AudioPlayer.cpp
+++ b/src/AudioPlayer.cpp
@@ -432,10 +432,8 @@ void AudioPlayer_Task(void *parameter) {
                     }
                     if (gPlayProperties.repeatCurrentTrack) { // End loop if button was pressed
                         gPlayProperties.repeatCurrentTrack = false;
-                        char rBuf[2];
-                        snprintf(rBuf, 2, "%u", AudioPlayer_GetRepeatMode());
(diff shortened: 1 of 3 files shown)


src/AudioPlayer.cpp
```diff
@@ -850,11 +832,6 @@ void AudioPlayer_TrackQueueDispatcher(const char *_itemToPlay, const uint32_t _l
     gPlayProperties.currentTrackNumber = _trackLastPlayed;
     char **musicFiles;
 
-    #ifdef MQTT_ENABLE
-        publishMqtt((char *) FPSTR(topicLedBrightnessState), 0, false);
```
fetzerch, line 854: @biologist79 I removed these 3 messages. It's not clear to me why the LED brightness or the playmode were sent here. If they were here for a good reason, we need to move them to the loop as well.

biologist79:
Picked by cherry-commit.
Thanks for your contribution!

PR #170, opened 2022-11-05
implemented repeated long button press

Xento:


biologist79:
Added by cherry-picking.
Thanks for you contribution!

PR #171, opened 2022-11-11
Add options for NEOpixel ring

strguntbr:
- LED_OFFSET: if defined must be in the range [0; NUM_LEDS-1]. The starting LED will be shifted by this number of pixels
I added this option because I wanted to compare the look when the ring starts at top to the ring starting at bottom without having to reassemble the box. I think this does not add much complexity and maybe someone else might need it

- PROGRESS_HUE_START and PROGRESS_HUE_END: defines the start and end hue for the multi LED progress indicator gradient. By default it starts at green and ends at red (just as it did before). However I did not like the fact that the volume is displayed the same way as the current progress. With this defines you can e.g. display the progess with a single color or use the whole rainbow.
I also slightly edited the calulcation for the volume gradient (because actually it was not going from 85 back to -5, but the actual end value depended on the number of LEDs. For 24 LEDs it was going back to -1) to make sure volume and progress gradients are exaxlty the same by default

- BUTTONS_LED: defines a GPIO that powers button (or any other kind of) LEDs. Those LEDs will be enabled by default and disabled when nightmode is set (or the brightness of Neopixels is manually set to a level below the level for nightmode)

### DIFF
--- a/src/Cmd.cpp
+++ b/src/Cmd.cpp
@@ -81,18 +81,13 @@ void Cmd_Action(const uint16_t mod) {
 				#ifdef MQTT_ENABLE
 					publishMqtt((char *) FPSTR(topicSleepTimerState), "0", false);
 				#endif
-				#ifdef NEOPIXEL_ENABLE
-					Led_ResetToInitialBrightness();
-				#endif
+				Led_ResetToInitialBrightness();
 			} else {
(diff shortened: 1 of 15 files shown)


biologist79:
Sorry for my late reply.
Will test within the upcoming days.

strguntbr:
No problem. I just saw there are conflicts. I will rebase and fix them

biologist79:
Added (finally) :-)
Thanks for your commit.

PR #174, opened 2022-11-27
Delay start of webstream, when wifi is not connected

DirkHeinke:
This is my first try to fix the start of webstream after deepsleep.

Translation of log messages and naming and placing of variables is not final. 

What do you think about this approach? Should I clean it up to make it ready for merge?


biologist79:
As described via PN: needs to be solved differently.
But I also do not have a perfect solution for this problem (to be honest).

PR #175, opened 2022-12-04
Bluetooth: show signal strength (RSSI) in sink mode

tueddy:
Just a small enhancement, show the bluetooth signal strength in sink mode.
Retested current implementation, both sink and source mode works fine. Removed free heap log, it's quite stable..

Grüße
Dirk


biologist79:
Thanks for your contribution!
Added it (with some codeformat-changes).

Issue #176, opened 2022-12-08
Increase readability + Refactoring


bliepp:
Hello guys, nice project. While diving into the code for a feature I wanted to add I found some parts to be very hard to read (at least for me). I broke it down to two aspects:
* Inconsistency when it comes to formatting (e.g. indentation in Led.cpp/Led.h is 4 spaces while in Log.cpp - but not Log.h - it is 2 spaces) and naming conventions). Also sometimes functions are declared where you wouldn't expect them (e.g. in Led.h there are declarations for AudioPlayer_GetInitVolume and AudioPlayer_SetInitVolume - twice) and function/variable handling could be improved.
* Sometimes heavy nesting/very long function bodies (OMG Led_Task ^^)

So as a result I would suggest to 
1. Introduce a coding styleguide and enforce it with something like [EditorConfig](https://editorconfig.org/) and
1. While applying it step by step also refactor the code into a more readable form (e.g. replacing the _Namespace_FunctionName_ style and using C++ specific syntax like namespaces or classes, using guard clauses, etc.)

Kindly
bliepp

biologist79:
Hi,

feel free to refactor + test :-)
I'm already spending by far too much time for ESPuino. So I won't.

bliepp:
Okay, my wife's gonna kill me but I'll try my best. Do you prefer one big PR refactoring big chunks or many small PR's refactoring refactoring module by module?

biologist79:
Regarding different indentions: was probably introduced because there've been working different people on it. Usually I take care that it looks the same but this is probably not the case everywhere.

LED is hard to understand - I know. There's also an idea to test another Neopixel-lib. But as I said: I' not willing to spend additional time. I'm currently working on PCB-stuff. And I guess this won't change in near future.


Don't know what's the best way to move to C++. Probably a good startpoint is to start a conversation at ESPuino's forum. Maybe others are interested, too.

fschrempf:
I just want to confirm that at least the code in the LED Task desperately needs some refactoring. It's barely readable at the moment. Thanks @bliepp for stepping up! I highly appreciate the effort!

bliepp:
> I just want to confirm that at least the code in the LED Task desperately needs some refactoring. It's barely readable at the moment. Thanks @bliepp for stepping up! I highly appreciate the effort!

This is actually the first thing I want to refactor as it seems to be the biggest load of them all. I'm just waiting for PR #171 to be merged to avoid merge conflicts.

[...]

PR #177, opened 2022-12-08
Enforce consistent coding style

bliepp:
Hey,

I asked in #176 for a consistent coding style, so I made this:

This PR introduces an `.editorconfig` file to enforce a consistent coding style (see [EditorConfig](https://editorconfig.org/), available as a VS Code plugin, but not needed). Also, most source code files (all except the settings files) are reformatted to fit this coding style. Especially the following changes were made:

* All files end with a single blank line. Trailing white spaces are trimmed.
* For C++ a tab of length 8 is used instead of 4 spaces. This visually increases the nesting depth making it easier to spot places which need to be refactoring.
* For HTML an indentation of length 4 is used to make the code more readable. Tabs are used to optimize for PROGMEM size.
* The `platformio.ini` file now uses 4 spaces for indentation consistently

My next step for upcoming pull request(s) would be a step-by-step refactoring of some code parts which would greatly benefit from a consistent coding style.

Please let me now if you prefer spaces over tabs, but for now I tried to loosely imitate the [Kernel coding style guide](https://www.kernel.org/doc/html/v4.10/process/coding-style.html). 

Kindly bliepp


teian:
Why only loosly imitate Kernel K&R and not simply stick to it fully? Was there a reason for it?

bliepp:
> Why only loosly imitate Kernel K&R and not simply stick to it fully? Was there a reason for it? Anyhow appreciate the work notheless

I'm planning on refactoring the whole code base step by step the next weeks and months. This first reformatting step was just introducing an editor enforced formatting using EditorConfig, where the options are limited. Also, K&E aims at plain C, while I plan to refactor using more C++ features. When working in refactoring I will stick more strictly to K&R as good as it is applicable to C++. For now this would be to much work for something that would get a refactor soon anyway.

PR #178, opened 2022-12-18
bugfix: visual errors on volume-control

Joe91:
two small visual bugs:
1. when changing the volume in idle, there was a short blinking of the LESs with the idle-colour.
2. when changing the volume in default mode the progress shines through for one frame.


biologist79:
I just merged https://github.com/biologist79/ESPuino/pull/184.
This one here can be closed, right?

fschrempf:
> This one here can be closed, right?

Yes, this PR should be obsolete now. @Joe91 Maybe you want to verify that the LED issue you observed are now fixed in master?

PR #181, opened 2022-12-19
feature: startup after switch from bluetooth

Joe91:
When switching back to normal from bluetooth because of an RFID-Tag this modification allows to start this tag directly after the reboot.


biologist79:
1) Ends up with endless errors at serial-logging if a user has not enabled PLAY_LAST_RFID_AFTER_REBOOT previously because there's no init in this case for the requested NVS-entry.

[...]
[E][Preferences.cpp:472] getString(): nvs_get_str len fail: lastRfid NOT_FOUND
[...]

2) Don't like the fact, that NVS.getString() is called at every cycle as I'd consider this somehow slow. Should be necessary to do this once.

Joe91:
Sorry, that was not inteded by me.
Why is this called cyclic? Or do you mean every boot-clycle? Will have anouther look at it and improve it. Thanks for the feedback.

biologist79:
Cycle was wrong wording by myself - sorry.
What I wanted to say: it is called within loop(). A way to avoid this in every loop is to use a status-flag (bool). Like so: https://github.com/biologist79/ESPuino/blob/master/src/main.cpp#L61.

Joe91:
Ah found what you mean. Really stupid mistake by me to overlook this. Is there a way to find out if a value already exists in the config?

src/main.cpp
```diff
@@ -105,6 +107,28 @@ bool testSPIRAM(void) {
 	}
 #endif
 
+// Get next RFID-tag if already set
+void loadNextRfid(void) {
+	if (recoverRfidFromBluetooth) {
```
fschrempf, line 112: Instead of nesting the whole code block in an if condition, better do:

```c
if (!recoverRfidFromBluetooth) {
    return;
}
recoverRfidFromBluetooth = false; // only try it once per bootup
[...]
```

fschrempf (changes requested):

src/main.cpp
Joe91: Much nicer 👍

fschrempf: @Joe91 It's your PR, so I would suggest you do it :wink:

Joe91:
Ist mir mittlerweile eher egal, daher schließe ich das Mal hier...

PR #186, opened 2022-12-26
Additional settings for button locking

fschrempf:
This includes two new settings related to locking of the controls/buttons:

* Add setting to lock controls by default
* Add setting to exclude rotary encoder from locking

### DIFF
--- a/src/RotaryEncoder.cpp
+++ b/src/RotaryEncoder.cpp
@@ -41,11 +41,13 @@ void RotaryEncoder_Readjust(void) {
 // Handles volume directed by rotary encoder
 void RotaryEncoder_Cyclic(void) {
 	#ifdef USEROTARY_ENABLE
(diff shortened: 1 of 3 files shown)


fschrempf:
On second thought I wonder if having unlocking the buttons via button itself is disabled on purpose? Is unlocking only intended to be done via modification tag?

biologist79:
Locking and unlocking both indeed are only intended to be used by rfid-mod and/or mqtt. However, one can assign CMD_LOCK_BUTTONS_MOD to a button but when doing so there's no way back if pressed, hrhr.

fschrempf:
> Locking and unlocking both indeed are only intended to be used by rfid-mod and/or mqtt. However, one can assign CMD_LOCK_BUTTONS_MOD to a button but when doing so there's no way back if pressed, hrhr.

Makes sense! I dropped the first commit from the PR.

biologist79:
Added.
Thanks for your contribution.

PR #187, opened 2022-12-27
Migrate compile-time configuration to Kconfig

fschrempf:
Kconfig (the Linux kernel's build configuration system) is a widespread tool to do build configuration. It can be easily integrated in the PlatformIO build of ESPuino and has some major advantages compared to the manual editing of preprocessor macros in header files:

* Comfortable and less error-prone configuration via `menuconfig`
* Makes adding new configuration options easier
* Clean separation of user-facing configuration options and internal macros
* Provides a good way to deduplicate code in the configuration headers
* Automatic generation of config header files via `genconfig`

@biologist79 This is an incomplete draft to show how the migration to Kconfig could look like. If you agree with the approach, I can add further commits to migrate the rest of the settings.

See https://forum.espuino.de/t/konfiguration-via-kconfig/1608/10?u=fschrempf for discussion and details.

Todo:

* [x] Add basic Kconfig support
* [x] Add workaround for defines that reference identifiers like other defines/enums/types/etc.
* [x] Add integration for PlatformIO config
* [x] Migrate Settings
* [x] Migrate Board Settings
* [ ] Add documentation
* [ ] Adjust CI
* [ ] Rebase on latest dev
* [ ] Test/Review


fschrempf:
This needs reworking to work with the current version that uses the esp-idf build tooling (already uses Kconfig) instead of the Arduino build tooling (Arduino is now used as a component).

Closing for now.

PR #188, opened 2022-12-28
Button.cpp: Fix headphone detection while controls are locked

fschrempf:
When controls are locked Button_Cyclic() is aborted and Port_Cyclic() never gets called. This means we don't get the correct status of the HP_DETECT signal.

To fix this abort Button_Cyclic() after Port_Cyclic() is called, so we can still detect if the headphones are connected.


biologist79:
Good point!
Thank you.

PR #195, opened 2023-01-28
Add a setting to spawn a password protected access point

IngoMeyer441:
This PR adds the possibility to secure the WiFi in access point mode with a hard coded password. If no password is given (the default), the WiFi is open. Thus, the change is backwards compatible.


biologist79:
Thanks for your contribution!

PR #196, opened 2023-01-28
Always allow to apply an RFID tag twice when the playlist is finished

IngoMeyer441:
If the macro `DONT_ACCEPT_SAME_RFID_TWICE_ENABLE` is defined, the same RFID card cannot be accepted again even after the playlist was finished. This PR allows to use the same RFID card again in this case.


biologist79:
Thanks for your contribution!

tml89:
What is the difference to: https://github.com/biologist79/ESPuino/pull/192 ?

biologist79:
https://forum.espuino.de/t/neues-feature-pausieren-wenn-rfid-karte-entfernt-wurde/541/11

tml89:
get it, thanks!

PR #198, opened 2023-02-02
Add support for control LEDs to be connected to the NeoPixel

mzanetti:
This commit allows to add additional WS2812 LEDs to be connected at the end of the NeoPixel ring.

This may be useful to lighten up other artwork, buttons, texts etc on the ESPuino. This feature is somewhat redundant to the already BUTTON_LEDS feature but it allows to use WS2812 LEDs that can be dimmed with the overall brightness of the neopixel leds, and with that also follow the night mode brightness instead of just going 100% or off like the BUTTON_LEDS feature which only turns on/off a GPIO. Also colors of the static LEDs can be configured through settings.h by defining them in the CONTROL_LEDS_COLORS array.

### DIFF
--- a/src/settings.h
+++ b/src/settings.h
@@ -209,8 +209,9 @@
 	//#################### Settings for optional Modules##############################
 	// (optinal) Neopixel
 	#ifdef NEOPIXEL_ENABLE
-		#define NUM_INDICATOR_LEDS		24          	// number of LEDs
-		#define NUM_STATIC_LEDS			0				// not used currently
+		#define NUM_INDICATOR_LEDS		24          	// number of Neopixel LEDs (formerly NUM_LEDS)
+		#define NUM_CONTROL_LEDS		0		// optional control leds (https://forum.espuino.de/t/statische-ws2812-leds/1703)
+                #define CONTROL_LEDS_COLORS		{}		// Colors for the control LEDs. Make sure it lists at least NUM_CONTROL_LEDS colors, e.g. for three control LEDs define: CONTROL_LEDS_COLORS {CRGB::Yellow, CRGB::Blue, 0xFFFFFF} (predefined colors: http://fastled.io/docs/3.1/struct_c_r_g_b.html)
 		#define CHIPSET					WS2812B     	// type of Neopixel
 		#define COLOR_ORDER				GRB
 		#define NUM_LEDS_IDLE_DOTS		4           	// count of LEDs, which are shown when Idle
--- a/src/Led.cpp
+++ b/src/Led.cpp
@@ -44,8 +44,9 @@
 	static uint8_t Led_NightBrightness = LED_INITIAL_NIGHT_BRIGHTNESS;
 	constexpr uint8_t Led_IdleDotDistance = NUM_INDICATOR_LEDS / NUM_LEDS_IDLE_DOTS;
 
-	static CRGBArray<NUM_INDICATOR_LEDS + NUM_STATIC_LEDS> leds;
+	static CRGBArray<NUM_INDICATOR_LEDS + NUM_CONTROL_LEDS> leds;
 	static CRGBSet indicator(leds(0, NUM_INDICATOR_LEDS - 1));
+	static CRGBSet controlLeds(leds(NUM_INDICATOR_LEDS, NUM_INDICATOR_LEDS + NUM_CONTROL_LEDS - 1));
 
 	TaskHandle_t Led_TaskHandle;
 	static void Led_Task(void *parameter);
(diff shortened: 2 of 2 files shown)


mzanetti:
image
Here's an example of what I used this for.

laszloh:
Hi,
I really like the idea of the static LEDs. I have one proposal for improvement of the code though.

Instead of replacing all calls to Fastled.show() and in the new function setting the static LEDs all the time, you could use CRGBSet ([reference](https://github.com/FastLED/FastLED/wiki/RGBSet-Reference) & [code](https://github.com/FastLED/FastLED/blob/master/src/pixelset.h)). This class allows to a create sub arrays of LEDs, which then can be accessed just like a native c array:

with these you could do something like:

```cpp
static CRGBArray<NUM_LEDS + STATIC_LEDS> leds;    // a bit smarter than a native c array
static CRGBSet indicator(leds, 0, NUM_LEDS-1);
static CRGBSet buttons(leds, NUM_LEDS, NUM_LEDS+STATIC_LEDS-1);

// initialization just like a native array
FastLed.addLeds</*.....*/>(leds, leds.size());

// [some code]

// update only the button leds
for(int i=0;i<STATIC_LEDS;i++)
    buttons[i] = CRGB::Blue + i;
FastLed.show();   // show the whole array

// update only the indicator
indicator = CRGB::Black;     // clear array by writing all pixels black
uint8_t hue = 0;
for(uint8_t i=0;i<indicator.size();i+=2) {
    // color every 2nd pixel
    indicator[i] = CHSV(hue, 255, 255);
    hue += 10;
}
FastLed.show();   // show the whole array
```

playground: https://wokwi.com/projects/356105063667527681


With this you could update the buttons only once and then "forget" about them while the led indicator does it's own thing.

One thing to remember is though, that you can not call FastLed.clear() since it would clear the underlying array and thus all separate views. To clear one sub-array, you can just write `view = CRGB::Black;` which fills the whole array with black.

mzanetti:
Hi, thanks for the feedback. I'm happy that you think this is a useful addition.

So, initially that's exactly what I wanted to do as I agree it's a bit odd to hook into every single Fastled.show() call. However, the problem is that the original implementation of Led_Task() does in fact call Fastled.clear() in various occasions which would then also clear my buttons. This led me to the code that always sets the static leds in every run. It seemed less error pronoe to me than hooking extra code for the static buttons into every place where Fastled.clear() is called. Anyhow, I don't have a strong opinion on it and am happy to go for whichever solution you prefer.

Given this Led_AddStatic() is now called quite often, I thought it'd be easier to the caller to just pass the original `leds` variable instead of also having to remember to use a different variable for that. Also, the next box I will build, will very likely have even more static leds to light up other artwork, not just buttons. That said, I'm happy to iterate further on it based on your feedback.

Joe91:
Looks much nicer this way, but maybe we can just replace the "fastLed.clear()" with a new funciton that only sets the "non-static-LEDs" to black. That way you don't have to think about the static-leds in every future animation/modification...
Really not sure what's the best solution there. Maybe others have an other opinion on this?

mzanetti:
> Hi, I've made the changes to move the indicators to CRGBSet & CRGBArray and created a Pull Request: #200
> 
> This should create a good basis for this PR. With this, we do not need to hook into FastLED.show().

Great rework! That helped. Branch is now based on https://github.com/biologist79/ESPuino/pull/200

tueddy:
Preparation for this feature is merged in DEV branch, thank's for contribution @laszloh !

As suggested in the forum, please consider making the name "STATIC" a little broader to reflect future control functions e.g. single Pause/Play LED

mzanetti:
Rebased the branch to dev and adjusted as discussed.

tueddy:
Changes are very moderate, that's fine!

I've pickup changes here https://github.com/biologist79/ESPuino/pull/198/files and defined for my 16er Neopixel 
NUM_INDICATOR_LEDS 12
NUM_CONTROL_LEDS 4

Indidcator is adjusted to 12 correctly but the 4 control LED leaves blank.
Seems like `Led_SetButtonLedsEnabled()` never been called?

src/settings.h
```diff
@@ -210,7 +210,8 @@
 	// (optinal) Neopixel
 	#ifdef NEOPIXEL_ENABLE
 		#define NUM_INDICATOR_LEDS		24          	// number of LEDs
-		#define NUM_STATIC_LEDS			0				// not used currently
+		#define NUM_CONTROL_LEDS		5		// number of control LEDs
```
tueddy, line 213: Please disable this feature by default (for existing users) and add some minimal hints, e.g.:

`#define NUM_INDICATOR_LEDS		24          	// number of Neopixel LEDs (formerly NUM_LEDS)
#define NUM_CONTROL_LEDS    	0				// optional control leds (https://forum.espuino.de/t/statische-ws2812-leds/1703)
#define CONTROL_LEDS_COLORS		{} 				// Colors for the control LEDs. Make sure it lists at least NUM_CONTROL_LEDS colors, e.g. for two control LEDs define: CONTROL_LEDS_COLORS {CRGB::Yellow, CRGB::Blue}
`

Sorry for bad formatting here..

grafik

mzanetti: Done. Also added a reference to the documentation for the available predefined CRGB colors and added a hint that plain hex codes could be used too.

PR #200, opened 2023-02-15
Move FastLED access to CRGBArray and CRGBSets

laszloh:
Change over to CRGBArray from native arrays and add a CRGBSet for the indicator. 

This allows to update the indicator independent of other parts of the CRBG array.

### DIFF
--- a/src/settings.h
+++ b/src/settings.h
@@ -202,9 +202,10 @@
 	//#################### Settings for optional Modules##############################
 	// (optinal) Neopixel
 	#ifdef NEOPIXEL_ENABLE
-		#define NUM_LEDS			24          	// number of LEDs
-		#define CHIPSET				WS2812B     	// type of Neopixel
-		#define COLOR_ORDER			GRB
+		#define NUM_INDICATOR_LEDS		24          	// number of LEDs
+		#define NUM_STATIC_LEDS			0				// not used currently
+		#define CHIPSET					WS2812B     	// type of Neopixel
+		#define COLOR_ORDER				GRB
 		#define NUM_LEDS_IDLE_DOTS		4           	// count of LEDs, which are shown when Idle
 		#define OFFSET_PAUSE_LEDS		false		// if true the pause-leds are centered in the mid of the LED-Strip
 		#define PROGRESS_HUE_START		85          	// Start and end hue of mulitple-LED progress indicator. Hue ranges from basically 0 - 255, but you can also set numbers outside this range to get the desired effect (e.g. 85-215 will go from green to purple via blue, 341-215 start and end at exactly the same color but go from green to purple via yellow and red)
--- a/src/Led.cpp
+++ b/src/Led.cpp
@@ -24,8 +24,8 @@
 
 	#ifndef LED_OFFSET
 		#define LED_OFFSET 0
-	#elif LED_OFFSET < 0 || LED_OFFSET >= NUM_LEDS
-		#error LED_OFFSET must be between 0 and NUM_LEDS-1
+	#elif LED_OFFSET < 0 || LED_OFFSET >= NUM_INDICATOR_LEDS
+		#error LED_OFFSET must be between 0 and NUM_INDICATOR_LEDS-1
 	#endif
 
 	// Time in milliseconds the volume indicator is visible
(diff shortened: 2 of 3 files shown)


biologist79:
Bug: in case a track is over and the playlist ends, all leds keep enlighted.

laszloh:
Found the culprit. This code snippet tries to access the indicator[NUM_LED]. CRGSet does not have any boundary checks implemented, so it allows access to elements outside fo the array (like c does too). So we overwrite some memory regio after the array and this crashes FastLED internally (I could not find out what) and the task just stops running. 

https://github.com/biologist79/ESPuino/blob/f16ece2a95e0f59be9bc64ee317b4eccaddcceb4/src/Led.cpp#L772-L774

The problem appears when `gPlayProperties.currentRelPos` gets > 98.99. Unfortunately the map function does not restrict the range of the return value, so if the input boundary goes outside the defined range the output follows.

f.e. with this values NUM_LEDS = 12, DIMMABLE_STATES = 50 I can get the following result from map:

```
[ 200644 ]  pos: 97.58 value: 594 full: 11 last 44
[ 200655 ]  pos: 97.58 value: 594 full: 11 last 44
[...]
[ 201514 ]  pos: 97.84 value: 594 full: 11 last 44
[ 201528 ]  pos: 97.84 value: 594 full: 11 last 44
[ 201539 ]  pos: 97.84 value: 594 full: 11 last 44
[ 201550 ]  pos: 98.05 value: 600 full: 12 last 0
[ 201562 ]  pos: 98.05 value: 600 full: 12 last 0
[...]
[ 203512 ]  pos: 98.99 value: 600 full: 12 last 0
[ 203523 ]  pos: 98.99 value: 600 full: 12 last 0
[ 203535 ]  pos: 98.99 value: 600 full: 12 last 0
[ 203546 ]  pos: 98.99 value: 600 full: 12 last 0
[ 203557 ]  pos: 98.99 value: 600 full: 12 last 0
[ 203581 ]  pos: 99.06 value: 606 full: 12 last 6
[ 208050 ]  info        : Closing audio file
```
ledValue=606 is bad, because that means fullLeds=12 & lastLed=6. With this, we then try to set the led with index 12 (so the 13th) to a value.

One thing I can not say, why this was not an issue before. Propably beacuse we were overwriting a native c array and we got lucky and only overwrote a variable not often used (or one, which is written before read, like `nextAnimaiton`).

mzanetti:
FWIW, I'm running this branch (and my static LED one on top) on my device for a few days now and I noticed that sometimes all the LEDs seem to get stuck on red and not refreshing any more at all until restarted. Now I'm not sure if it's this branch or the previous rework one which is already merged (likely the previous already merged one) and also I don't know how to reproduce yet. But I've seen it in this state twice now already after my son used the device.

laszloh:
Were you running the latest commit here [9dbf039](https://github.com/biologist79/ESPuino/pull/200/commits/9dbf039b0b29cadfa71463bb40ff84f5d47c6a20)(or at least [4180920](https://github.com/biologist79/ESPuino/pull/200/commits/4180920d35c4203ad3ca36b9a682b219dbae0685))? Because without these, the playback progress is buggy (accessing out of bound elements of the LED array).

laszloh:
I'll push this PR back to draft version, until I update it to the current dev branch

mzanetti:
> Were you running the latest commit here [9dbf039](https://github.com/biologist79/ESPuino/pull/200/commits/9dbf039b0b29cadfa71463bb40ff84f5d47c6a20)(or at least [4180920](https://github.com/biologist79/ESPuino/pull/200/commits/4180920d35c4203ad3ca36b9a682b219dbae0685))? Because without these, the playback progress is buggy (accessing out of bound elements of the LED array).

I've updated to the latest state a couple of days ago, and I could not see that behavior any more since.

laszloh:
So I've finished the rebase onto dev.

src/settings.h
```diff
@@ -202,9 +202,10 @@
 	//#################### Settings for optional Modules##############################
 	// (optinal) Neopixel
 	#ifdef NEOPIXEL_ENABLE
-		#define NUM_LEDS			24          	// number of LEDs
-		#define CHIPSET				WS2812B     	// type of Neopixel
-		#define COLOR_ORDER			GRB
+		#define NUM_INDICATOR_LEDS		24          	// number of LEDs
+		#define NUM_STATIC_LEDS			0				// not used currently
```
fschrempf, line 206: What's the purpose of `NUM_STATIC_LEDS`?

laszloh: A remainder from my test codes to check, that I do not change anything expect the CRGBSet. 
I left the define in to show, how additional LEDs can be added to the LED array, f.e. #200.

fschrempf (approved):
LGTM!

laszloh:
This is also finished from my side.

mzanetti:
I'll rebase mine (#198) once this is merged.

tueddy:
Hi, i think this is a good base for supporting additional static LEDs beside the existing indicator LEDs in future.
I can''t see any breaking changes, LGTM!

PR #201, opened 2023-02-21
Impove BT-Source device scanning

tueddy:
- use device scanning callback & show available devices in Log
- With empty "nameBluetoothSourceDevice" BT-Source connects to first device found 
- Update A2DP lib to latest version


biologist79:
Merged!
Thanks for your support.

PR #209, opened 2023-03-18
Bugfix: Corrupted files after HTTP-upload

tueddy:
Bugfix: playlistcache was invalid after "Rename"


tueddy:
As discussed in https://forum.espuino.de/t/probleme-mit-mp3-dateien/1671

PR #218, opened 2023-03-25
Platformio.ini Refractoring and Bugfix

laszloh:
This commit moves the common flags (like monitor_filter and build_flags) to the global env in platformio. 
This allows the central maintanance of the build_flags and reduce the number of copy & paste errors.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -15,12 +15,17 @@ default_envs = lolin_d32_pro_sdmmc_pe
 [env]
 board_build.flash_mode = qio
 board_build.bootloader = dio
+board_build.partitions = custom_4mb_noota.csv
 ;platform = espressif32@<=3.5.0
-platform = espressif32@<=6.1.0
+platform = espressif32@^6.1.0
 ;platform = espressif32
 ;platform = https://github.com/platformio/platform-espressif32.git#feature/arduino-upstream
 framework = arduino
 monitor_speed = 115200
+monitor_filters = esp32_exception_decoder
+;monitor_port = /dev/cu.SLAB_USBtoUART
+;upload_port = /dev/cu.SLAB_USBtoUART
+;upload_speed = 115200
 extra_scripts =
     pre:gitVersion.py
     pre:processHtml.py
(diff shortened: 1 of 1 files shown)


platformio.ini
```diff
@@ -16,6 +16,11 @@ default_envs = lolin_d32_pro_sdmmc_pe
 board_build.flash_mode = qio
 board_build.bootloader = dio
 board_build.partitions = custom_4mb_noota.csv
+
+; performance changes, enable these to make the esp faster (also expect higher power usage)
+board_build.f_cpu = 240000000L
```
fschrempf, line 21: All the board configurations we currently support already have `f_cpu` set to 240 MHz  by default. What's the purpose of duplicating that here?

platformio.ini
```diff
@@ -16,6 +16,11 @@ default_envs = lolin_d32_pro_sdmmc_pe
...
+;board_build.f_flash = 80000000L
```
fschrempf, line 22: The commit message claims that the flash clock is set to 80 MHz, but the line is commented here. Anyway I don't think we should override the default settings from the board configurations by default. The value from the board config should already provide a sane default.

fschrempf (changes requested):

laszloh:
ok, I'll split it in two PR's. This one for the fixing of the Arduino verison and introduction of global build flags and a new one to transistion to c++17.

platformio.ini
laszloh: Funnily I have not seen, that there is a whole list of boards and their default CPU speed. I'll update the comment and comment it out so that users can activate it, if they want to.

platformio.ini
laszloh: I'll update the commit message, I tried it with 80MHz for a short time, but it did not show any improvements.

fschrempf:
> ok, I'll split it in two PR's. This one for the fixing of the Arduino verison and introduction of global build flags and a new one to transistion to c++17.

Ok, two PRs wouldn't really be strictly necessary in my opinion. Separate commits in the same PR would also have been fine. I just want to make sure that the Git history is somewhat meaningful after merging.

PR #223, opened 2023-04-03
Fix for snprintf trunctation warnings

laszloh:
This PR adds the following functionality / changes:

- a printf wrapper to the LogBuffer system. The max log length is 200 bytes and "..." is added to the end of the string, if it was truncated. 
- Add missing translations text's
- Move the formatting of the log strings into the translation. This enabled language specific position of the variables.
- Fix compiler warnings
  - To fix truncation warning, the functionality of `Web_DeleteCachefile` had to be changed. The function now takes a folder and deletes the playlist cache file in it.
  - Do not use [strncpy](https://cplusplus.com/reference/cstring/strncpy/), since no null-character is copied at the end of the string, if source is longer than destination.
  - Remove comparison warnings which evaluate to always true


src/Log.cpp
```diff
@@ -51,6 +51,28 @@ void Log_Print(const char *_logBuffer, const uint8_t _minLogLevel, bool printTim
...
 
+int Log_Printf(const uint8_t _minLogLevel, const char *format, ...) {
+	char loc_buf[201];	// Allow a maximum buffer of 200 characters in a single log message
+
+	int len;
+	va_list arg;
+	va_start(arg, format);
+    
+    // use the local buffer and trunctate string if it's larger
+    len = vsnprintf(loc_buf, sizeof(loc_buf), format, arg);
```
fschrempf, line 57: Indentation should use tabs instead of spaces.

src/Log.cpp
```diff
@@ -51,6 +51,28 @@ void Log_Print(const char *_logBuffer, const uint8_t _minLogLevel, bool printTim
...
+
+	Log_Print(loc_buf, _minLogLevel, true);
+	if(len > sizeof(loc_buf) - 1) {
+		// long string was trunctated
+		Log_Print("...", _minLogLevel, false);
+	}
+	Log_Print("\n", _minLogLevel, false);
+
+    va_end(arg);
```
fschrempf, line 66: Indentation should use tabs instead of spaces.

src/Web.cpp
```diff
@@ -350,11 +351,13 @@ void webserverStart(void) {
 			request->redirect("https://espuino.de/espuino/favicon.ico");
 		});
         // Init HallEffectSensor Value
-        #ifdef HALLEFFECT_SENSOR_ENABLE
+        #ifndef HALLEFFECT_SENSOR_ENABLE
```
fschrempf: Why are you changing the preprocessor logic here?

src/AudioPlayer.cpp
```diff
@@ -823,8 +823,11 @@ void AudioPlayer_TrackQueueDispatcher(const char *_itemToPlay, const uint32_t _l
 		}
 	#endif
 	char filename[255];
+    
+    size_t sizeCpy = strnlen(_itemToPlay, sizeof(filename) - 1);    // get the len of the play item (to a max of 254 chars)
+    memcpy(filename, _itemToPlay, sizeCpy);
+    filename[sizeCpy] = '\0';       // terminate the string
```
fschrempf, line 829: Indentation  wrong

fschrempf (changes requested):
Looks good to me except for some small issues I commented.

src/Web.cpp
fschrempf: You are reverting this change in a later commit. Would be good if you could remove it from commit https://github.com/biologist79/ESPuino/pull/223/commits/475ba3776af09705898bcb25928c2f4a5e0a3a7d in the first place.

src/Log.cpp
laszloh: fixed

src/Log.cpp
laszloh: fixed

src/Web.cpp
laszloh: yes, I had changed the ifdef to test the code and missed it in the first commit. It's now fixed.

src/AudioPlayer.cpp
laszloh: done

PR #229, opened 2023-04-16
Remove PROGMEM and supporting macros

laszloh:
We do not need to use the makros PGMSPACE or PSTR(), since [pgmspace.h](https://github.com/espressif/arduino-esp32/blob/master/cores/esp32/pgmspace.h) is an empty helper header to make Arduino Libraries not specifically written for the esp32 work.

This code does not aim to support other platforms (like ESP8266, where pgmspace.h is used), so we do not need to use the makros.

This PR also fixes the typecasts like `( (char*) FPSTR(mqttOk) )` which loose the const modifier.
` FPSTR(mqttOk)` becomes `(reinterpret_cast<const __FlashStringHelper *>(mqttOk))` after makro expansion (see [WString.h](https://github.com/espressif/arduino-esp32/blob/master/cores/esp32/WString.h) row 37). This is then typecast to `(char*)` which could lead to undefined behavior if the caller expects a mutable char array.


fschrempf:
Thanks! Looks good to me. Is there any specific reason why this is still a draft?

fschrempf (approved):
LGTM!

laszloh:
I've rebased it to the current HEAD of the dev branch. This PR is ready from my side.

tueddy:
@laszloh Unfortunately, there is a new conflict with the last commit. Can you resolve it?
One question: Apart from the clean-up changes, are there any advantages in memory use/consumption?

PR #230, opened 2023-04-16
Small bugfixes

tueddy:
- fix crash playing a .Mp3/.mP3 files (make file extension really case-insensitive). .mp3, MP3 was OK, .Mp3, .mP3 crashed
- fix reading enhanced M3U playlist with comment lines ("#EXTINF"), see https://forum.espuino.de/t/crash-beim-2-abspielen-einer-m3u-playlist/1921
- ~~fix mem-leak in loop openNextFile(), close to free file object~~
- add own icon for playlists in Web-ui
- JavaScript .test() is case-insensitive, remove unnecessary comparison
- add all audio formats supported by the ESP32-audioI2S library (.opus, .pls, .asx)
- Stop playing before explorer contextmenu DELETE (can crash if deleted file is playing currently)

I did not found a better icon in free font-awesome for a playlist:
grafik

I hope codestyle/formatting is OK..

### DIFF
--- a/src/SdCard.cpp
+++ b/src/SdCard.cpp
@@ -105,11 +105,14 @@ bool fileValid(const char *_fileItem) {
 	// make file extension to lowercase (compare case insenstive)
 	char *lFileItem;
 	lFileItem = x_strdup(_fileItem);
+	if (lFileItem == NULL) {
+		return false;
+	}
 	lFileItem = strlwr(lFileItem);
 	const char ch = '/';
 	char *subst;
 	subst = strrchr(lFileItem, ch); // Don't use files that start with .
-	return (!startsWith(subst, (char *) "/.")) && (
+	bool isValid = (!startsWith(subst, (char *) "/.")) && (
 			// audio file formats
 			endsWith(lFileItem, ".mp3") || 
 			endsWith(lFileItem, ".aac") || 
@@ -123,6 +126,8 @@ bool fileValid(const char *_fileItem) {
 			endsWith(lFileItem, ".m3u8") || 
 			endsWith(lFileItem, ".pls") || 
 			endsWith(lFileItem, ".asx"));
+	free(lFileItem);
+	return isValid;
 }
 
 
--- a/src/SdCard.cpp
+++ b/src/SdCard.cpp
@@ -160,7 +160,6 @@ char *SdCard_pickRandomSubdirectory(char *_directory) {
 			break;
 		}
 		if (!fileItem.isDirectory()) {
-			fileItem.close();
 			continue;
 		} else {
 			#if ESP_ARDUINO_VERSION_MAJOR >= 2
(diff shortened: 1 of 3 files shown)


fschrempf:
It would be really good if you could do one commit per change/fix. It saves a lot oft time when reviewing the PR, or debugging the code. Thanks!

tueddy:
@fschrempf OK next time. This PR is 90% case-insensitive comparision and should be straight forward & easy to understand.
If you have time to review:

I'm a bit unsure if this is memory-safe or could be optimized further:
https://github.com/tueddy/ESPuino/blob/cfcf85004563200be80f095b6f8d38a068aeb3c1/src/SdCard.cpp#L107

laszloh:
> I'm a bit unsure if this is memory-safe or could be optimized further: https://github.com/tueddy/ESPuino/blob/cfcf85004563200be80f095b6f8d38a068aeb3c1/src/SdCard.cpp#L107

No, this is **not** memory-safe. strdup (and x_strdup) duplicates the string with malloc, so you must free it after usage. 
But you do not need a strdup here:
1. use a local char array of f.e. 32 elements (but at least max. extension + 2) created on the stack
2. find position of last dot
3. check if string fits into local array --> if not return false
4. if yes, copy extension into local array
5. make it lower case
6. check against array of extensions

For the array of valid extensions I'd propose the following:
```cpp
static constexpr std::array<const char*> extensions[]={
    ".mp3",
    ".m3u",
    ".wav",
    ".flac",
    [...]
};
[...]
for(const auto &ext: extensions) {
    if(endsWith(lFileItem, ext)){
        return true;
    }
}
// no extensions matched
return false;
```

src/SdCard.cpp
```diff
@@ -146,14 +155,15 @@ char *SdCard_pickRandomSubdirectory(char *_directory) {
...
+			fileItem.close();
 			continue;
 		} else {
 			#if ESP_ARDUINO_VERSION_MAJOR >= 2
 				strncpy(buffer, (char *) fileItem.path(), 255);
 			#else
 				strncpy(buffer, (char *) fileItem.name(), 255);
 			#endif
-
+			fileItem.close();
```
laszloh, line 171: It's not necessary to call `fileItem.close()` here. `fileItem` is a local variable, so it's destructor is called when it get's out of scope (this is at the end of the loop).
The File Object has a `std::shared_ptr<FileImpl> p` holding a pointer to the file system object. As soon as the File object is destroyed, the smart pointer has 0 references to it and it in turn calls the destructor of the object it holds. This calls the virtual destructor `FileImpl::~FileImpl()` and this is forwarded to the implementation `VFSFileImpl ::~VFSFileImpl()` which calls close on the file object [here](https://github.com/espressif/arduino-esp32/blob/7693f7410a1885f8345be377b8197f60b5cd93f6/libraries/FS/src/vfs_api.cpp#L345).

src/Web.cpp
```diff
@@ -1067,6 +1067,10 @@ void explorerHandleDeleteRequest(AsyncWebServerRequest *request) {
 		param = request->getParam("path");
 		convertFilenameToAscii(param->value(), filePath);
 		if (gFSystem.exists(filePath)) {
+			// stop playback, file to delete might be in use
+			if (!gPlayProperties.pausePlay) {
```
laszloh, line 1071: I assume that if we are paused, delete the file and then resume the playback, that we also get a crash. So here you should rather check, if we are in playback mode (not only if we are not paused). 

For a fast fix, this is ok, but it'd also be more elegant to check, if we are deleting a file/folder from the folder which is currently playing, before stopping the music playback (not sure how complicated this would be to implement).

laszloh:
look good

PR #231, opened 2023-04-16
HallEffectSensor files:

NIKO-AT:
Source beautified with CLang format


biologist79:
Partially merged via cherry-picking.

PR #233, opened 2023-04-23
FS speedup by factor > 20 (playlist generation, Web-UI file listing)

tueddy:
As discussed here: https://forum.espuino.de/t/beschleunigte-verzeichnisauflistung-bei-sehr-vielen-mp3s/1348 

To see the difference in speed, i pressed relaod in browser many times:

Current:

https://user-images.githubusercontent.com/11274319/233864879-af5e9c9b-8334-47e0-8cd7-921e00db3493.MOV

With this PR:

https://user-images.githubusercontent.com/11274319/233864865-34046806-9e4e-483b-9ec9-53eb204d3d49.MOV

### DIFF
--- a/src/SdCard.cpp
+++ b/src/SdCard.cpp
@@ -180,7 +180,6 @@ char *SdCard_pickRandomSubdirectory(char *_directory) {
 			#else
 				strncpy(buffer, (char *) fileItem.name(), 255);
 			#endif
-			fileItem.close();
 		#endif
 
 			// Log_Printf(LOGLEVEL_INFO, nameOfFileFound, buffer);
@@ -332,6 +331,11 @@ char **SdCard_ReturnPlaylist(const char *fileName, const uint32_t _playMode) {
 			serializedPlaylist[0] = '#';
 			while (fileOrDirectory.available() > 0) {
 				buf = fileOrDirectory.read();
+				if (buf == '#') {
+					// skip M3U comment lines starting with #
+					fileOrDirectory.readStringUntil('\n');
+					continue;
+				}				
 				if (fPos+1 >= allocCount * allocSize) {
 					char *tmp = (char *) realloc(serializedPlaylist, ++allocCount * allocSize);
 					Log_Println((char *) FPSTR(reallocCalled), LOGLEVEL_DEBUG);
@@ -424,7 +428,6 @@ char **SdCard_ReturnPlaylist(const char *fileName, const uint32_t _playMode) {
 				#else
 					strncpy(fileNameBuf, (char *) fileItem.name(), sizeof(fileNameBuf) / sizeof(fileNameBuf[0]));
 				#endif
-				fileItem.close();
 			#endif
 				// Don't support filenames that start with "." and only allow .mp3 and other supported audio file formats
 				if (fileValid(fileNameBuf)) {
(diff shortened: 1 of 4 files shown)


src/SdCard.cpp
```diff
@@ -388,19 +405,31 @@ char **SdCard_ReturnPlaylist(const char *fileName, const uint32_t _playMode) {
...
 
 		while (true) {
-			File fileItem = fileOrDirectory.openNextFile();
-			if (!fileItem) {
-				break;
-			}
-			if (fileItem.isDirectory()) {
-				continue;
-			} else {
+			#if defined(HAS_FILEEXPLORER_SPEEDUP) || (ESP_ARDUINO_VERSION_MAJOR == 2 && ESP_ARDUINO_VERSION_MINOR == 0 && ESP_ARDUINO_VERSION_PATCH >= 8)
```
laszloh, line 408: See comment in SdCard_pickRandomSubdirectory
```suggestion
			#if defined(HAS_FILEEXPLORER_SPEEDUP) || ESP_ARDUINO_VERSION >= ESP_ARDUINO_VERSION_VAL(2, 0, 8)
```

src/Web.cpp
```diff
@@ -925,39 +925,56 @@ void explorerHandleListRequest(AsyncWebServerRequest *request) {
 		return;
 	}
 
-	File file = root.openNextFile();
+	#if defined(HAS_FILEEXPLORER_SPEEDUP) || (ESP_ARDUINO_VERSION_MAJOR == 2 && ESP_ARDUINO_VERSION_MINOR == 0 && ESP_ARDUINO_VERSION_PATCH >= 8)
```
laszloh, line 928: Also here, if you want to use the combined macro (it should be consistently used)
```suggestion
	#if defined(HAS_FILEEXPLORER_SPEEDUP) || ESP_ARDUINO_VERSION >= ESP_ARDUINO_VERSION_VAL(2, 0, 8)
```

src/Web.cpp
```diff
@@ -925,39 +925,56 @@ void explorerHandleListRequest(AsyncWebServerRequest *request) {
...
-		if (!startsWith( file.path() , (char *)"/.")) {
-		#else
-		if (!startsWith( file.name() , (char *)"/.")) {
-		#endif
-			JsonObject entry = obj.createNestedObject();
+		while (file) {
+			// ignore hidden folders, e.g. MacOS spotlight files
 			#if ESP_ARDUINO_VERSION_MAJOR >= 2
-				convertAsciiToUtf8(file.path(), filePath);
+				if (!startsWith( file.path() , (char *)"/.")) {
```
laszloh, line 949: This block could be made more readable by using a const variable (instead of fetching the file path twice) and then you can use the same code as in the loop in Line 934:

```cpp
			// ignore hidden folders, e.g. MacOS spotlight files
			const String MyfileName;
			#if ESP_ARDUINO_VERSION_MAJOR >= 2
				MyfileName = file.path();
			#else
				MyfileName = file.name();
			#endif
			if(!startsWith(MyfileName, "/:")) {
				JsonObject entry = obj.createNestedObject();
				convertAsciiToUtf8(MyfileName.c_str(), filePath);
				// and so on
			}
```

laszloh (changes requested):
One other point, you seem to have the PR not based on the current dev branch and have a random commit in it: [d34da44](https://github.com/biologist79/ESPuino/pull/233/commits/d34da44b29fefdfb6103ac6343f6e9567d2108d0)

PR #235, opened 2023-04-30
(re-)allow to compile without any RFID-reader

tueddy:
All non-essential tasks such as audio player, RFID and LED are paused to speed up the web upload.
Until now, a global task handle was used for this. However, this no longer compiles if no RFID-reader is selected. This could be the case for testing or special applications such as a pure web player.
This PR allows compiling even without an RFID reader and does not need global declarations. It also clear all LEDs instead of just freezing..


fschrempf (approved):
LGTM!

Before you merge, can you squash the two commits into one, please? Thanks!

PR #241, opened 2023-06-01
use bssid to connect to best WIFI and added option to always perform a WIFI-scan

Joe91:
Option to alsways perform a WIFI-Scan.
Always connect to the strongest possible WIFI after a scan.

### DIFF
--- a/src/Wlan.cpp
+++ b/src/Wlan.cpp
@@ -125,7 +125,7 @@ void Wlan_Init(void) {
 	handleWifiStateInit();
 }
 
-void connectToKnownNetwork(WiFiSettings settings) {
+void connectToKnownNetwork(WiFiSettings settings, byte *bssid = nullptr) {
 	// set hostname on connect, because when resetting wifi config elsewhere it could be reset
 	if (hostname.compareTo("-1")) {
 		WiFi.setHostname(hostname.c_str());
@@ -145,7 +145,7 @@ void connectToKnownNetwork(WiFiSettings settings) {
 
 	Log_Printf(LOGLEVEL_NOTICE, wifiConnectionInProgress, settings.ssid);
 
-	WiFi.begin(settings.ssid, settings.password);
+	WiFi.begin(settings.ssid, settings.password, 0, bssid);
 }
 
 void handleWifiStateInit() {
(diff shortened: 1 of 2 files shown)


src/Wlan.cpp
```diff
@@ -160,7 +160,12 @@ void handleWifiStateInit() {
 	connectionAttemptCounter = 0;
 	connectStartTimestamp = 0;
 	connectionFailedTimestamp = 0;
-	wifiState = WIFI_STATE_CONNECT_LAST;
+	#ifdef ALWAYS_SCAN_WIFI_ON_STARTUP
```
tueddy, line 163: As discussed please use a NVS setting here, e.g. 
`bool performScan = gPrefsSettings.getBool("ALWAYS_SCAN_WIFI_ON_STARTUP", true);`

src/settings.h
```diff
@@ -56,6 +56,7 @@
 	//#define SAVE_PLAYPOS_BEFORE_SHUTDOWN  // When playback is active and mode audiobook was selected, last play-position is saved automatically when shutdown is initiated
 	//#define SAVE_PLAYPOS_WHEN_RFID_CHANGE // When playback is active and mode audiobook was selected, last play-position is saved automatically for old playlist when new RFID-tag is applied
 	//#define HALLEFFECT_SENSOR_ENABLE      // Support for hallsensor. For fine-tuning please adjust HallEffectSensor.h Please note: only user-support provided (https://forum.espuino.de/t/magnetische-hockey-tags/1449/35)
+	//#define ALWAYS_SCAN_WIFI_ON_STARTUP	// Slows down first connection but ensures to always connect to the best WIFI
```
tueddy, line 59: This would be obsolet if we use a NVS setting

src/Wlan.cpp
Joe91: Will do so.
Where can we modify this option afterwards? Shall we add a checkbox in the wifi-settings in the Webserver for that or would you just leave it that way for now?

tueddy: I would preset performScan with true, so at default ESPuino startup performs a scan and always connects to stronger WiFi (same name).

We could add this setting later in Web-UI. At this moment we can live without it ;-)
Just my opinion..

PR #246, opened 2023-06-16
Stricter hostname validation

SZenglein:
Changes the frontend validation and adds a backend validation.
The backend previously did not validate the hostname at all.

The hostname validation is primarily for mDNS. For the "pretty" hostname
much more could be allowed, but we do not make that distinction (yet).

In general, change what kind of hostnames are accepted.
For example, a hostname can have a '-' in the middle.
Also, only alphanumerical characters are allowed. Dots ('.') are still
disallowed to prevent DNS problems when using mDNS.

### DIFF
--- a/html/accesspoint.html
+++ b/html/accesspoint.html
@@ -22,6 +22,9 @@
 				width: 20px;
 				height: 20px
 			}
+			input:invalid {
+				border: red solid 3px;
+			}
 			body {
 				background: #007bff;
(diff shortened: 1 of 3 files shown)


tueddy:
merged by cherry pick in https://github.com/biologist79/ESPuino/commit/dc8d7eeace3928d393dd1fb1ae597da5f03a6a82
Thank's for your contribution!

PR #252, opened 2023-06-25
Feature/modification: shuffle playlist

h4kun4m4t4t4:
this PR adds the possibility to shuffle a running playlist via action **CMD_SHUFFLE_PLAYLIST** either via card or webUI.
After applying the action the current title will be stopped, the playlist will be randomized and started immediately afterwards.

This way one can bind favorites songs to Card A and (if bored 😉) randomize the order of tracks played with Card B.
[Wikipedia: Shuffle Mode](https://en.wikipedia.org/wiki/Shuffle_play)


biologist79:
Will not be added. Thanks anyway for your contribution.

PR #255, opened 2023-07-10
Update .gitignore

JanHBade:
ignore more files from VS Code


fschrempf:
Can you please give a reason for this change? Tracking some oft the VSCode configuration files is intended and useful. I don't see a reason for ignoring them. Please add custom ignores to a local untracked ignore file within your workspace or user namespace.

Further you are ignoring files that are still in the repo. If you ignore them, you should also remove them from the repo.

JanHBade:
grafik

the change are comes and comes but i don't change that file.... so I think ist is a generated file that should not tracked
(the same for the settins.json)

yes i forgot to delete.

fschrempf:
> the change are comes and comes but i don't change that file

I have no idea why this is the case for you, but I don't see this issue here. I don't see any changes in the screenshot.

> so I think ist is a generated file that should not tracked (the same for the settins.json)

No, I think it should be tracked as is.

tueddy:
The aim of this PR is to prevent the files e.g. ".vscode/settings.json" from appearing in Github Source control view. Because they have not been changed by the user and are not relevant.
As far as I can see settings.json is important to tell VSCode to use PlatformIO. Anyone have a better solution to keep these files in the repository and still exclude it in the source control view?

JanHBade:
grafik

this window is gone if you delete the .vscode folder, I dont think we need this window, we have a great documention ;)

tueddy:
I have re-checked in extensions.json but it makes no differencce, file is already shown as modified.
Hiding the dialog about installing PlatformIO cannot be a solution for new users.

My suggestion is to just ignore it and live with it.

fschrempf:
@JanHBade @tueddy Can you tell me the exact steps to reproduce the problem, please? I don't see it here. Is this only in VSCode built into GitHub (Codespaces)? Is it also a problem in local VSCode?

tueddy:
I can reproduce the change in VSCode with Win11, maybe it's on windows only. Strange, the file seems to be unchanged (diff with Notepad+)

JanHBade:
the problem is away (maybe newest version from vscode make it...)

PR #262, opened 2023-09-13
bugfix_same_rfid_twice init

Joe91:
Found a small issue:
when using the function "DONT_ACCEPT_SAME_RFID_TWICE" it only works the 2nd time because the Audioplayer resets this function until it starts playing.

Additionally I would suggest to also remove Line 47 in RfidCommon.cpp.
https://github.com/biologist79/ESPuino/blob/dev/src/RfidCommon.cpp#L47
In my opinion there is no need to accept unused cards as "last used" cards since they have no function yet. This prevents wrong UID-Readouts (that can happen with a not so ideal distance to the reader) from messing with the current audio playback.
What's your opinion on that?

### DIFF
--- a/src/AudioPlayer.cpp
+++ b/src/AudioPlayer.cpp
@@ -817,8 +817,14 @@ void AudioPlayer_Task(void *parameter) {
 		//esp_task_wdt_reset(); // Don't forget to feed the dog!
 
 		#ifdef DONT_ACCEPT_SAME_RFID_TWICE_ENABLE
+			static uint8_t resetOnNextIdle = false;
 			if (gPlayProperties.playlistFinished || gPlayProperties.playMode == NO_PLAYLIST) {
-				strncpy(gOldRfidTagId, "X", cardIdStringSize-1);     // Set old rfid-id to crap in order to allow to re-apply an rfid-tag after playback is finished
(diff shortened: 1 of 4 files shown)


pitlochry:
> Additionally I would suggest to also remove Line 47 in RfidCommon.cpp.
https://github.com/biologist79/ESPuino/blob/dev/src/RfidCommon.cpp#L47
In my opinion there is no need to accept unused cards as "last used" cards since they have no function yet. This prevents wrong UID-Readouts (that can happen with a not so ideal distance to the reader) from messing with the current audio playback.
What's your opinion on that?

I would strongly suggest to remove this line. It is now line 50 in RfidCommon.cpp (lines 49 to 51 can be removed). Otherwise applying an unused card (which does not have a function yet) results into accepting the same rfid twice, which should be excluded by DONT_ACCEPT_SAME_RFID_TWICE, e.g.:
Card1 -> Unused Card (which does not do anything .. so card1 still plays) -> Card1 (is accepted)

Expected behaviour (achieved by removing the line):
Card1 -> Unused Card (which does not do anything .. so card1 still plays) -> Card1 (is **not** accepted, since it is the same as already applied)

tueddy:
removed as suggested..

PR #263, opened 2023-10-18
fix build for EN language

pcppcp:
this fixes some minor typos when -DLANGUAGE=EN is used


biologist79:
Merged!
Thanks for your contribution.

PR #269, opened 2023-11-21
Playlist improvements

laszloh:
Modifies the playlist generation to use C++ std::vector and std::string with a custom allocator (preferring the PSRAM if available) to create the file list.

For an extended discussion about the features and class hierarchy, see: https://forum.espuino.de/t/vorstellung-playlist-optimierung/1890/11


src/Common.h
```diff
@@ -22,6 +24,14 @@ inline bool isNumber(const char *str) {
 	}
 }
 
+inline const char *getPath(File &f) {
+	if constexpr (ESP_ARDUINO_VERSION_MAJOR >= 2) {
```
tueddy, line 28: This is obsolete, we are on Arduino >=2

PR #270, opened 2023-11-21
CardMan: Autoload Files from Card/Cloud

QDaniel:
https://forum.espuino.de/t/cloud-anbindung/721/19

### DIFF
--- a/src/CardMan.cpp
+++ b/src/CardMan.cpp
@@ -1,0 +1,235 @@
+
+#include "CardMan.h"
+#include "Log.h"
+#include "base64url.h"
+#include <HTTPClient.h>
+#include "SdCard.h"
+#include "AudioPlayer.h"
+
+#define CHUNK_SIZE      1024
+
+ 
+cardmanData cardData;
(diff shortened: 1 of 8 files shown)


tueddy:
What you are submitting here is so specific, I don't think it makes sense for the vast majority of users

- We have generally decided against writing to a card. There are also tags that will no longer work after writing on other known jukeboxes
- You have only implemented it for RC522 but not for PN5180
- Logging is outside the ESPuino standard methods

QDaniel:
- The Tags without writing still work
- Current is it only for RC522 i don't have a PN5180 for testing
- Logging must be corrected

tueddy:
Which cloud service do you actually want to connect to? Please provide a detailed description here or in the forum. I didn't even understand your intention for this PR despite the forum topic...

The majority of users use the PN5180, so it must at least be tested against both readers.  The reader costs < 10 € and should be available for you.

Once again: We have rejected writing to an RFID card because it can break reading the tags for other jukeboxes

PR #274, opened 2023-12-04
Immediately go to sleep if battery is critical

SZenglein:
Explanation is given in the code.

### DIFF
--- a/src/Battery.cpp
+++ b/src/Battery.cpp
@@ -30,7 +30,11 @@ void Battery_Init(void) {
 
 		Log_Println(batteryCriticalMsg, LOGLEVEL_NOTICE);
 		// Power down and enter deepsleep
-		System_RequestSleep();
+		// Don't call System_RequestSleep() here: If the battery is critial, we want to avoid as much init work as possible
+		// and also any blinking lights or sounds. The goal is to just stay off.
+		// Additionally, LPCD will not be enabled. This is intentional to avoid battery drain.
+		delay(200);
+		esp_deep_sleep_start();
 	}
 	#endif
 }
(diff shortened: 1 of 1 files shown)


laszloh:
You should take a look at [System_DeepSleepManager](https://github.com/biologist79/ESPuino/blob/8407213b5a2d9ddff69430b294b8700acb39a680/src/System.cpp#L212) to see what actions are taken before entering the deep sleep.

What I see a is, that you do not turn of any peripheral units, so both the AMP and the power system could still be active . I'd at least propose to call the two functions ``System_PreparePowerDown`` and  ``Power_PeripheralOff`` (this one is the more important one) to bring the two GPIO pins to the required level, so the peripherals are unpowered.

SZenglein:
System_PreparePowerDown is a good pointer. 

The thing is, Battery_Init is intentionally placed very early in the setup function, meaning the peripheral power has not yet been turned on. Also Audioplayer has not been initialized, so it should not be exited either.

It has its quirks and order of operations in setup is a difficult topic, but I did design the battery module, the power module and influenced the setup order with all that in mind, and as far as I can see it has not changed.

Essentially, it's treated in the same way as esp_deep_sleep_start in Pn5180 when a false LPCD wakeup ha been detected. 


Also, I think you might have missed this because I only wrote it in a private message to @tueddy, but this PR is essentially just reverting this hidden change: https://github.com/biologist79/ESPuino/commit/63dc6954686bc2c4d885a3ed5bd1af09e8ceb5fd#diff-f983eb1c0981492d0daafbef571290cd784c5b9cec1526ea3f344f7729e81232L31

laszloh:
Yes, I did not knew the history of this change. In the light of it, it looks good.

PR #278, opened 2023-12-09
fix wrong states on PE output pins (and SD-card failure on restart)

r-schmidt:
The output register status buffer was initialized fix to {0xff,0xff} although the output registers were set to {0x00,0x00}. That buffer is used by Port_Write() to determine the values of the bits for the other pins. So those get set to 1 (high).

Since the peripheral power on the Espuino 4-layer mini PCB is switched on by setting a pin on the PCA9555 low, that power is first turned on (by Port_Init()), then turned off (by the Port_Write() call in AudioPlayer_Init()) and then turned on again (by Power_PeripheralOn()). That caused a 12.5 millisecond disruption of the external 3.3V which some SD-cards don't like.

### DIFF
--- a/src/Port.cpp
+++ b/src/Port.cpp
@@ -20,7 +20,7 @@
 extern TwoWire i2cBusTwo;
 
 uint8_t Port_ExpanderPortsInputChannelStatus[2];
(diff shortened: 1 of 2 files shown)


tueddy:
Tested with Lolin D32 pro /SD-MMC /blue PCB and Mini4Layer PCB, everything works as expected!
Applied by cherry-pick to DEV branch. Thanks for your contribution!

PR #279, opened 2023-12-10
Long press behaviour

Joe91:
Execute Longpress directly when intervall reached...


tueddy:
For reference:
https://forum.espuino.de/t/probleme-mit-cmd-prevtrack/2509

PR #280, opened 2023-12-12
/rfid endpoint - Serve with chunked response

tueddy:
/rfid endpoint:

A static buffer of 8KB is currently used to list the RFID tags. If this buffer is full, the returned JSON is truncated and not all entries are displayed in the web interface. . This happens with me from approx. 60 entries (depends on path length). It could also lead to a low memory situation and, in the worst case, to a crash.

With this PR, a list is first created holding the keys only and later the details such as path and game mode are sent as a chunked response with a smaller buffer. 
No changes in the delivered JSON except that it is always complete

### DIFF
--- a/src/Web.cpp
+++ b/src/Web.cpp
@@ -1753,7 +1753,7 @@ static void handleGetRFIDRequest(AsyncWebServerRequest *request) {
 					Log_Println("/rfid: Buffer too small", LOGLEVEL_ERROR);
 					return 0;
 				}
-				len += snprintf(((char *) buffer), maxLen-len, "[%s", json.c_str());
+				len += snprintf(((char *) buffer), maxLen - len, "[%s", json.c_str());
 				nvsIndex++;
 			}
 			while (nvsIndex < nvsKeys.size()) {
@@ -1762,12 +1762,12 @@ static void handleGetRFIDRequest(AsyncWebServerRequest *request) {
 				if ((len + json.length()) >= maxLen) {
 					break;
 				}
-				len += snprintf(((char *) buffer + len), maxLen-len, ",%s", json.c_str());
+				len += snprintf(((char *) buffer + len), maxLen - len, ",%s", json.c_str());
 				nvsIndex++;
 			}
 			if (nvsIndex == nvsKeys.size()) {
 				// finish
-				len += snprintf(((char *) buffer + len), maxLen-len, "]");
+				len += snprintf(((char *) buffer + len), maxLen - len, "]");
 				nvsIndex++;
 			}
 			return len;
--- a/src/Web.cpp
+++ b/src/Web.cpp
@@ -1751,7 +1751,7 @@ static void handleGetRFIDRequest(AsyncWebServerRequest *request) {
 				json = tagIdToJsonStr(nvsKeys[nvsIndex].c_str(), idsOnly);
 				if (json.length() >= maxLen) {
 					Log_Println("/rfid: Buffer too small", LOGLEVEL_ERROR);
-					return 0;
+					return len;
 				}
 				len += snprintf(((char *) buffer), maxLen - len, "[%s", json.c_str());
 				nvsIndex++;
--- a/src/Web.cpp
+++ b/src/Web.cpp
@@ -457,7 +457,7 @@ void webserverStart(void) {
 
 		// RFID
 		wServer.on("/rfid", HTTP_GET, handleGetRFIDRequest);
-		wServer.addRewrite(new OneParamRewrite("/rfid/ids", "/rfid?ids=true"));
+		wServer.addRewrite(new OneParamRewrite("/rfid/ids-only", "/rfid?ids-only=true"));
 		wServer.addHandler(new AsyncCallbackJsonWebHandler("/rfid", handlePostRFIDRequest));
 		wServer.addRewrite(new OneParamRewrite("/rfid/{id}", "/rfid?id={id}"));
 		wServer.on("/rfid", HTTP_DELETE, handleDeleteRFIDRequest);
@@ -1709,8 +1709,8 @@ static String tagIdToJsonStr(const char *key, const bool nameOnly) {
 }
 
 // Handles rfid-assignments requests (GET)
-// /rfid returns an array of tag-id keys
-// /rfid/details returns an array of tag-ids and details. Optional GET param "id" to list only a single assignment.
+// /rfid returns an array of tag-ids and details. Optional GET param "id" to list only a single assignment.
+// /rfid/ids-only returns an array of tag-id keys
 static void handleGetRFIDRequest(AsyncWebServerRequest *request) {
 
 	String tagId = "";
@@ -1726,7 +1726,7 @@ static void handleGetRFIDRequest(AsyncWebServerRequest *request) {
 		return;
 	}
 	// get tag details or just an array of id's
-	bool idsOnly = request->hasParam("ids");
+	bool idsOnly = request->hasParam("ids-only");
 
 	std::vector<String> nvsKeys {};
 	static size_t nvsIndex;
--- a/src/Web.cpp
+++ b/src/Web.cpp
@@ -1729,7 +1729,7 @@ static void handleGetRFIDRequest(AsyncWebServerRequest *request) {
 	bool idsOnly = request->hasParam("ids");
 
 	std::vector<String> nvsKeys {};
-	static uint16_t nvsIndex;
+	static size_t nvsIndex;
 	nvsKeys.clear();
 	// Dumps all RFID-keys from NVS into key array
 	listNVSKeys("rfidTags", &nvsKeys, DumpNvsToArrayCallback);
@@ -1741,7 +1741,7 @@ static void handleGetRFIDRequest(AsyncWebServerRequest *request) {
 	// construct chunked repsonse
 	nvsIndex = 0;
 	AsyncWebServerResponse *response = request->beginChunkedResponse("application/json",
-		[nvsKeys = std::move(nvsKeys), idsOnly](uint8_t *buffer, size_t maxLen, size_t index) {
+		[nvsKeys = std::move(nvsKeys), idsOnly](uint8_t *buffer, size_t maxLen, size_t index) -> size_t {
 			maxLen = maxLen >> 1; // some sort of bug with actual size available, reduce the len
 			size_t len = 0;
 			String json;
@@ -1772,7 +1772,6 @@ static void handleGetRFIDRequest(AsyncWebServerRequest *request) {
 			}
 			return len;
 		});
-	nvsKeys.clear();
 	request->send(response);
 }
 
--- a/src/Web.cpp
+++ b/src/Web.cpp
@@ -1685,34 +1685,81 @@ bool DumpNvsToJSONCallback(const char *key, void *data) {
 	return tagIdToJSON(key, obj);
 }
 
+// callback for writing a NVS entry to list
+bool DumpNvsToArrayCallback(const char *key, void *data) {
+	std::vector<String> *keys = (std::vector<String> *) data;
+	keys->push_back(key);
+	return true;
+}
+
+static String tagIdToJsonStr(const char *key) {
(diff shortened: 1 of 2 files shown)


src/Web.cpp
```diff
@@ -1685,34 +1685,81 @@ bool DumpNvsToJSONCallback(const char *key, void *data) {
...
+				json = tagIdToJsonStr(nvsKeys[nvsIndex].c_str());
+				if ((len + json.length()) > maxLen) {
+					break;
+				}
+				len += sprintf(((char *) buffer + len), ",%s", json.c_str());
+				nvsIndex++;
+			}
+			if (nvsIndex == nvsKeys.size()) {
+				// finish
+				len += sprintf(((char *) buffer + len), "]");
```
SZenglein, line 1757: Edge case: that single byte for ']' or ',' could be one byte more than maxLen. 

Why not write '[' and ']' by themselves without sprintf?

src/Web.cpp
```diff
@@ -1685,34 +1685,81 @@ bool DumpNvsToJSONCallback(const char *key, void *data) {
...
 
+// callback for writing a NVS entry to list
+bool DumpNvsToArrayCallback(const char *key, void *data) {
+	std::vector<String> *keys = (std::vector<String> *) data;
+	keys->push_back(key);
+	return true;
+}
+
+static String tagIdToJsonStr(const char *key) {
+	StaticJsonDocument<512> doc;
```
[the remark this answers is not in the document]
tueddy: A single tag id looks like:

`{
        "id": "003108198106",
        "fileOrUrl": "/Die Schönsten Lieder Zum Einschlafen",
        "playMode": 3,
        "lastPlayPos": 1842837,
        "trackLastPlayed": 0
    }`

maxLen of "fileOrUrl" is MAX_FILEPATH_LENTGH = 256, size calculation with https://arduinojson.org/v6/assistant gives about 350 Bytes needed. Recommended is power of 2, so 512 Bytes is best size.

src/Web.cpp
```diff
@@ -1685,34 +1685,81 @@ bool DumpNvsToJSONCallback(const char *key, void *data) {
...
-			request->send(404);
-		}
+
+	if ((tagId != "") && gPrefsRfid.isKey(tagId.c_str())) {
+		// return single RFID entry
+		json = tagIdToJsonStr(tagId.c_str());
+		request->send(200, "application/json", json);
+		return;
 	}
+	static std::vector<String> nvsKeys {};
```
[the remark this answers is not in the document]
tueddy: nvsKeys is declared as local variable and i assume it gets out of scope after function handleGetRFIDRequest() finished.
Am i wrong here?

src/Web.cpp
```diff
@@ -1685,34 +1685,81 @@ bool DumpNvsToJSONCallback(const char *key, void *data) {
...
+	AsyncWebServerResponse *response = request->beginChunkedResponse("application/json",
+		[](uint8_t *buffer, size_t maxLen, size_t index) {
+			maxLen = maxLen >> 1; // some sort of bug with actual size available, reduce the len
+			size_t len = 0;
+			static String json;
+
+			if (nvsIndex == 0) {
+				// start, write first tag
+				json = tagIdToJsonStr(nvsKeys[nvsIndex].c_str());
+				len += sprintf(((char *) buffer), "[%s", json.c_str());
```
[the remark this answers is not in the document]
tueddy: Good catch! 
I never saw a buffer smaller than 1500 Bytes so JSON length never exceeds.
But i will do an additional check here for code safety.

src/Web.cpp
```diff
@@ -1685,34 +1685,81 @@ bool DumpNvsToJSONCallback(const char *key, void *data) {
...
+		json = tagIdToJsonStr(tagId.c_str());
+		request->send(200, "application/json", json);
+		return;
 	}
+	static std::vector<String> nvsKeys {};
+
+	nvsKeys.clear();
+	// Dumps all RFID-keys from NVS into key array
+	listNVSKeys("rfidTags", &nvsKeys, DumpNvsToArrayCallback);
+	if (nvsKeys.size() == 0) {
```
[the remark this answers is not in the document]
tueddy: Writing the first tag there is no comma in JSON, it is starting with the second entry.
That's is the reason to handle the special case size==0 here and not even start a chunked response.

src/Web.cpp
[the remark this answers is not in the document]
SZenglein: It's declared as static, meaning it lives as long as the entire program.
If you didn't add nvsKeys.clear() directly afterwards, I it should continue to grow with each call.

The upside is that push_back does not need to reserve more memory every time and instead can re-use the same heap space. The downside is that the heap space that is the capacity of the vector is never reduced or freed.

src/Web.cpp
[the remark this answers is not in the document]
SZenglein: Not starting the chunked response in the first place is probably a good idea.

src/Web.cpp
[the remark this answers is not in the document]
laszloh: > The downside is that the heap space that is the capacity of the vector is never reduced or freed.
And because of this, I'd advice against static here. The main issue is, that we do not have any idea, how many entries there are and we are basically reserving heap space (at the minimum `sizeof(String) * [maxEntries]`, propably way more).

Also what you have to keep in mind is, that as soon as std::vector runs out of space in it's internal array, it reserves a new memory block on the heap with the factor of 1.5 / 2 (so if an array of 12 elements is full and another object is pushed, the vector will reserve max 24 entries before moving the memory). This is to ensure an O(n) for reallocations over lifetime of the object.

src/Web.cpp
```diff
@@ -1685,34 +1685,81 @@ bool DumpNvsToJSONCallback(const char *key, void *data) {
...
+	JsonObject entry = doc.createNestedObject(key);
+	if (!tagIdToJSON(key, entry)) {
+		return "";
+	}
+	String serializedJsonString;
+	serializeJson(entry, serializedJsonString);
+	return serializedJsonString;
+}
+
+uint16_t nvsIndex;
```
laszloh, line 1706: Move `nvsIndex` either as a static variable into the lambda in line 1736 or at least make the variable `static` (so that it does not pollute the global namespace).

src/Web.cpp
```diff
@@ -1685,34 +1685,81 @@ bool DumpNvsToJSONCallback(const char *key, void *data) {
...
+	listNVSKeys("rfidTags", &nvsKeys, DumpNvsToArrayCallback);
+	if (nvsKeys.size() == 0) {
+		// no entries
+		request->send(200, "application/json", "[]");
+		return;
+	}
+	// construct chunked repsonse
+	nvsIndex = 0;
+	AsyncWebServerResponse *response = request->beginChunkedResponse("application/json",
+		[](uint8_t *buffer, size_t maxLen, size_t index) {
```
laszloh, line 1735: If `nvsKeys` is not static any more, we have to capture the variable here so it's available in the lambda. To prevent a copy of the vector due to capture by value, we call std::move to transfer the variable into the lambda (capture by reference would result in undefined behaviour since the local variable will not exists any more when we are called).

```suggestion
		[nvsKeys = std::move(nvsKeys)](uint8_t *buffer, size_t maxLen, size_t index) {
```

src/Web.cpp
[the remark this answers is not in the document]
laszloh: This looks ok for me, we do not start the chunked response here (just send back an empty array and return)

src/Web.cpp
laszloh: +1 for not using sprintf but directly accessing the array

src/Web.cpp
```diff
@@ -457,6 +457,7 @@ void webserverStart(void) {
 
 		// RFID
 		wServer.on("/rfid", HTTP_GET, handleGetRFIDRequest);
+		wServer.addRewrite(new OneParamRewrite("/rfid/details", "/rfid?details=true"));
```
SZenglein, line 460: OneParamRewrite is usually for path parameters in the form of /path/{some_param} -> /path?some_param={some_param} because AsyncWebServer cannot handle path params natively. 

Does this even work and do something? Even if it does, using the "normal" AsyncWebRewrite makes more sense: https://github.com/me-no-dev/ESPAsyncWebServer#param-rewrite-with-matching

src/Web.cpp
```diff
@@ -1732,33 +1742,38 @@ static void handleGetRFIDRequest(AsyncWebServerRequest *request) {
...
 			size_t len = 0;
-			static String json;
+			String json;
 
 			if (nvsIndex == 0) {
 				// start, write first tag
-				json = tagIdToJsonStr(nvsKeys[nvsIndex].c_str());
-				len += sprintf(((char *) buffer), "[%s", json.c_str());
+				json = tagIdToJsonStr(nvsKeys[nvsIndex].c_str(), withDetails);
+				if (json.length() > maxLen) {
```
SZenglein, line 1752: You check that, but 4 lines later you're writing json.length() + 1 bytes, so you should also check for json.length() + 1 available space

src/Web.cpp
```diff
@@ -1732,33 +1742,38 @@ static void handleGetRFIDRequest(AsyncWebServerRequest *request) {
...
+					return len;
+				}
+				len += snprintf(((char *) buffer), maxLen, "[%s", json.c_str());
 				nvsIndex++;
 			}
 			while (nvsIndex < nvsKeys.size()) {
 				// write tags as long we have enough room
-				json = tagIdToJsonStr(nvsKeys[nvsIndex].c_str());
+				json = tagIdToJsonStr(nvsKeys[nvsIndex].c_str(), withDetails);
 				if ((len + json.length()) > maxLen) {
```
SZenglein, line 1762: Same issue, check is for 1 byte too few because of the comma

src/Web.cpp
```diff
@@ -1732,33 +1742,38 @@ static void handleGetRFIDRequest(AsyncWebServerRequest *request) {
...
 					break;
 				}
-				len += sprintf(((char *) buffer + len), ",%s", json.c_str());
+				len += snprintf(((char *) buffer + len), maxLen, ",%s", json.c_str());
 				nvsIndex++;
 			}
 			if (nvsIndex == nvsKeys.size()) {
 				// finish
-				len += sprintf(((char *) buffer + len), "]");
+				len += snprintf(((char *) buffer + len), maxLen, "]");
```
SZenglein, line 1770: That byte is written without a length check at all.

SZenglein:
Regarding the buffer copying, it's still calling for memory issues IMHO.
Most importantly, every code like this:
```
snprintf(((char *) buffer + len), maxLen, "...", ....);
```
Needs to be converted to 
```
snprintf(((char *) buffer + len), maxLen-len, "]");
```
that would already prevent any invalid memory access when writing the buffer. If the string doesn't fit, the json would still be invalid though.

src/Web.cpp
tueddy: Yes it works fine, do you have a concrete improved code here?

SZenglein: Didn't test it, but just replacing 'OneParamRewrite' with 'AsyncWebRewrite' should do the same thing.

laszloh:
> Needs to be converted to `snprintf(((char *) buffer + len), maxLen-len, "...", ...);`

Yes you are correct. It's probably easier if we introduce a local variable for the remaining capacity `const size_t remaining = maxLen - len - 1;` (-1 for the last ']').

> If the string doesn't fit, the json would still be invalid though.

From my point of view still better than getting a buffer overflow, overwriting unknown memory blocks after the buffer and having at best an unhandled exception at the next malloc/free call. We can also check for an error from snprintf by `(len < 0 || len >= remaining)` and react accordingly (I'd need to look it up, if there is a possibility to abort a chunked response in ESPAsyncWebServer & return f.e. HTTP-Code 500).

laszloh (commented):
Looks good, I've added some changes to the code (mostly cosmetic). The code works as expected with entry count of over 300 dummy cards.

PR #282, opened 2023-12-26
improve port expander handling

r-schmidt:
This is a collection of changes I made for the port expander:

- prevent the PE ISR from doing anything if there was no real interrupt (no idea where those spurious interrupts come from)
- speed up access to the PE
- make the debouncer work per bit

### DIFF
--- a/src/Port.cpp
+++ b/src/Port.cpp
@@ -20,7 +20,7 @@
 extern TwoWire i2cBusTwo;
 
 uint8_t Port_ExpanderPortsInputChannelStatus[2];
-static uint8_t Port_ExpanderPortsOutputChannelStatus[2] = {255, 255}; // Stores current configuration of output-channels locally
+static uint8_t Port_ExpanderPortsOutputChannelStatus[2]; // Stores current configuration of output-channels locally
 void Port_ExpanderHandler(void);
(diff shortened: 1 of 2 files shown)


tueddy:
@r-schmidt  Thank you very much for your contribution!

Please do the PR in the DEV branch first if possible, we can still test the extensions intensively there.
Part of your extension is already included in the DEV branch, see here:
https://github.com/biologist79/ESPuino/commit/2ef1b07398ccd0f2d6ff008131d6f82d376b2cac

As discussed we use uint_8 not unsigned int in loop: 
https://forum.espuino.de/t/neustart-schlaegt-fehl-abhaengig-von-sd-karte/2460/31
 
Where exactly did you encounter problems with the current implementation? Perhaps you can briefly describe it in the forum so that everyone can get an idea of how you have done speedup & debounce

r-schmidt:
Seems like I need to make a new PR to change the base branch on my side.

tueddy:
@r-schmidt I have tested your code & there does seem to be an optimised & faster access to the PE (less CPU time is used). Only PE port 2 I could not test now. It's worth transferring the PR against the DEV branch, thanks for your effort!

tueddy:
Follow up https://github.com/biologist79/ESPuino/pull/283

PR #283, opened 2023-12-26
improve port expander handling

r-schmidt:
This is a collection of changes I made for the port expander:

__prevent the PE ISR from doing anything if there was no real interrupt__
I've seen the ISR being called all the time and thus Port_ExpanderHandler() accessing the bus every time it gets called. That wastes a lot of time which can be spent more useful. So I added a check in the ISR if there was really an interrupt request from the PE.

__correct interrupt mode for the PE__
The PCA9555 interrupt output is level low, not falling edge. That is not very different if there is only one device using the line but there is a small chance that input register 0 changes while reading input register 1. And in that case the interrupt line would stay low which won't be detected when it is set for falling edge.
So using ONLOW fixes that but requires that the interrupt is disabled in the ISR and re-enabled after reading the input registers.

__speed up access to the PE__
The current implementation isn't wrong, just horribly slow because to read 2 bytes from the PE it puts 3 unnecessary bytes on the bus plus i2c timing overhead.

__make the debouncer work per bit__
I didn't encounter an issue but if one of the bits in the input registers was changing all the time, all the other bits would be ignored. There shouldn't be a debouncer at all but until there is a thread safe i2c handling, that can't be avoided.

### DIFF
--- a/src/Port.cpp
+++ b/src/Port.cpp
@@ -38,13 +38,16 @@ void Port_Init(void) {
 #ifdef PORT_EXPANDER_ENABLE
 	Port_Test();
 	Port_WriteInitMaskForOutputChannels();
-	Port_ExpanderHandler();
 #endif
 
 #ifdef PE_INTERRUPT_PIN_ENABLE
 	pinMode(PE_INTERRUPT_PIN, INPUT_PULLUP);
-	attachInterrupt(PE_INTERRUPT_PIN, PORT_ExpanderISR, FALLING);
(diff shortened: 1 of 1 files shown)


tueddy:
I've tested against the complete profile, it uses 108 for headphone detect. It's also working fine.
So i can confirm port 0 + port 1 is fully functional. Overall also audio crackling problems are gone with your PR.
It seems to use far less CPU time.

108 => port 1 channel/bit 0

PR #294, opened 2024-02-02
improve progress calculation/indication

freddy36:
This improved the `gPlayProperties.currentRelPos` calculation by making it dependent on the availability of the file size instead of the `isWebstream` flag. Accuracy is also slightly improved by subtracting the metadata header size (`audio->getAudioDataStartPos()`) of the file. 
LED animations will also check for the available file size instead of the `isWebstream` flag.
This enables correct progress indication when streaming static files.


tueddy:
Thanks for your work, looks good!
Do you have an example webstream or is this generally valid?

tueddy:
The example made the problem very easy to understand. 
Thank you very much for your contribution!

PR #296, opened 2024-02-05
use natural sort

freddy36:
Currently media with >= 10 files won't be played in the correct order unless the filenames contain leading zeros.
This changes the playlist generation and the explorer file list sorting to natural sort.
It makes sure that the names containing numeric values are sorted by their actual value (1 < 2 < 10) instead of a basic string comparison (1 < 10 < 2).


tueddy:
Changing the sort order is a "breaking change" for existing users. 
It would also be nice to announce and discuss this beforehand in the forum with a topic.

A new library will be integrated. Have you compared the memory consumption?

freddy36:
A topic has been created here:
https://forum.espuino.de/t/umstellung-auf-natural-sort/2781

Issue #272, opened 2023-11-29
Remove convertAsciiToUtf8()


orempel:
Moin!

With the change to Arduino v2 "everything" is utf8 ready. Which means that the `convertAsciiToUtf8()` function is actually doing more harm that good:  playing some files result in constant errors in the terminal and no updates in the web interface:

```
I [21103] Kommando: Nächster Titel
I [21116] info        : Closing audio file
I [21131] info        : buffers freed, free Heap: 87180 bytes
I [21131] info        : Reading file: "/Giraffenaffen/Giraffenaffen 1/03-Hätt’ ich Dich heut’ erwartet (feat. Roman Lob).mp3"
I [21185] info        : MP3Decoder has been initialized, free Heap: 59360 bytes , free stack 2948 DWORDs
N [21189] '/Giraffenaffen/Giraffenaffen 1/03-Hätt’ ich Dich heut’ erwartet (feat. Roman Lob).mp3' wird abgespielt (3 von 17)D [21193] ws[/ws][1] error(1002): Invalid UTF-8 in text frame":3,"numberOﾭ�xV��K�?Q
D [21210] ws[/ws][1] disconnect
...
D [27104] ws[/ws][2] connect
D [27199] ws[/ws][2] error(1002): Invalid UTF-8 in text frame"],"active":ﾭ�xV����?0
D [27200] ws[/ws][2] disconnect
D [32940] ws[/ws][3] connect
D [33031] ws[/ws][3] error(1002): Invalid UTF-8 in text frame"],"active":ﾭ�xV����?0
D [33031] ws[/ws][3] disconnect
D [38878] ws[/ws][4] connect
D [38971] ws[/ws][4] error(1002): Invalid UTF-8 in text frame"],"active":ﾭ�xV����?�
D [38972] ws[/ws][4] disconnect
```

I will create a PR for that, but just want to ask if it make sense to rip out `convertFilenameToAscii()` as well. That function was updated for the utf8 change and is now just copying c-strings around.

Thanks!

tueddy:
`convertAsciiToUtf8()` is obsolete and can be removed. Copying arround was just for smooth porting to Arduino 2.
But i think removing does not fix `error(1002): Invalid UTF-8 in text frame"],"active":ﾭ�xV����?0` ?

orempel:
I've already checked that, removing `convertAsciiToUtf8()` fixes the problem.

[...]

PR #298, opened 2024-02-08
remove convertFilenameToAscii()

freddy36:
avoid unnecessary string copies by removing the legacy `convertFilenameToAscii()` function

Fixes memory corruption when working with long file paths (`convertFilenameToAscii()` didn't check the target buffer size).


tueddy:
This fixes second part of  https://github.com/biologist79/ESPuino/issues/272

PR #299, opened 2024-02-08
Playlist fixes

freddy36:
fix a crash when trying to play an "invalid" file (playlist with length 1 was created but without an actual item)

Avoid unecessary string copy during playlist generation


tueddy:
LGTM!
The playlist will probably be rebuilt shortly, so your fixes may no longer be necessary. However, a log output for invalid files makes sense.

PR #302, opened 2024-02-11
send numbers as json number instead of string

freddy36:
Use the + operator to convert strings representing numbers to real number types. Should make the JSON parsing on the ESPuino slightly more efficient.


laszloh:
I think it'd be better to use `Number()` or `parseInt` for converting. It's more expressive than the unary operator+ in what you are doing.

I fear that in the future the single + could be removed if it's intention is not clear. In contrast Number/parseInt explicitly state what is happening here.

freddy36:
changed it to use `Number()`

PR #314, opened 2024-04-13
Embed swagger UI

caco3:
This PR embeds the Swagger UI into the Web Interface. To keep the additionally required flash usage low, most files get pulled from a CDN.

I also renamed the last tab `Forum` to `Help` and extended the tab with a section for the REST API documentation.
The great advantage for an embedded Swagger UI is to be able to directly send the cmds to the device.

# Updated UI
grafik

# Swagger UI
grafik

# Direct calls to the device
grafik

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ sdkconfig.*
 src/settings-override.h
 src/settings-custom-override.h
 node_modules/
+html/REST_API.yaml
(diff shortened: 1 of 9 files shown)


tueddy:
Just tested, works fine!

tueddy:
@caco3  Looks good so far! 
One issue i found, Web.cpp is getting recompiled even nothing has changed:

```
Building in release mode
Compiling .pio\build\lolin_d32_pro_sdmmc_pe\src\Web.cpp.o
Linking .pio\build\lolin_d32_pro_sdmmc_pe\firmware.elf
```
We had this before and https://github.com/biologist79/ESPuino/pull/293 fixed the issue.
Maybe same here?

caco3:
Good point, I changed it to make sure the mtime does not change.

BTW, this PR adds 3544 bytes to the flash usage. If this is a problem, we could create a define in `settings.h` so it could be added optionally. The static RAM usage is unchanged although we added 3 extra files to the webserver.

tueddy:
Tested against Lolin D32 (without PSRAM) & additional flash usage is OK

PR #315, opened 2024-04-15
Added virtual RFID-Cards

caco3:
This PR adds 5 virtual RFID-Cards.

It is based on the concept in https://forum.espuino.de/t/webradio-station-auf-eine-taste-legen/1343 but for now kept as simple as possible:

1. In `settings.h` assign any button action to one of the virtual RFID-Card cmds `CMD_VIRTUAL_RFID_CARD_0{1..5}`
2. In the WebUI, use the corresponding virtual RFIDs `90000000000{1..5}` as the RFID number.

If the PR shall be accepted, I will also add some documentation to the WebUI.

Instead of using the numbers, the WebUI also could use a letter based RFID, eg. `VIRTUAL_RFID_0{1..5}`, how ever this would need extension of the parser in the WebUI and probably also in the firmware.

### DIFF
--- a/html/management.html
+++ b/html/management.html
@@ -37,9 +37,11 @@
 		}
 		.filetree {
 			border: 1px solid black;
-			height: 200px;
 			margin: 0em 0em 1em 0em;
 			overflow-y: scroll;
+		}
+		.filetree-size {
+			height: 200px;
 		}
 		.slider.slider-horizontal {
 			width: 60%;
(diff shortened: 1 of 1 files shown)


caco3:
This is a mock-up of the UI:
grafik

@tueddy @biologist79 Please let me know if you would accept it and then I will implement it properly in the UI

PR #321, opened 2024-04-20
[obsolete] Upgrade to Bootstrap 5, added dark mode switch

caco3:
- Upgrade to Bootstrap 5
- Added dark mode (persisted in local storage)
- fixed formating
- fixed json compression
- ~~convert yaml to json for size reduction~~ (Not working because of Github pipeline issues)
- minify all html/js/css files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,8 @@ src/settings-override.h
 src/settings-custom-override.h
 node_modules/
 html/REST_API.yaml
+html/*.min.html
+html/locales/*.min.json
+html/REST_API.min.yaml
+html/js/*.min.min.js
+html/js/swaggerInitializer.min.js
--- a/.gitignore
+++ b/.gitignore
@@ -16,8 +16,3 @@ src/settings-override.h
 src/settings-custom-override.h
 node_modules/
 html/REST_API.yaml
-html/*.min.html
-html/locales/*.min.json
-html/REST_API.min.yaml
-html/js/*.min.min.js
-html/js/swaggerInitializer.min.js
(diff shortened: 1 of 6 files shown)


fschrempf:
@caco3 Would you mind cleaning up the git history just a little bit? Otherwise we will get one big commit with different changes (if squashed during merge) or several nonsense fixup commits as now. Both options are somewhat messy and make it really hard to do debugging, etc. afterwards.

Just running a `git rebase` and squashing some fixups and edit some commit messages should do the job. Thanks!

fschrempf:
> Here, playing with UI stuff, I tend to have the need to play around and do commits just as a backup as it is mostly try&error.

Playing around and try and error is totally fine. For the future it might be helpful to use `git commit --fixup=...`. It doesn't cost that much time and helps you to clean up afterwards.

> Usually, when I try to do a rebase (to dev), it usually messes it up, so I am trying to avoid it.  But with some help we could have a look on it.

Rebasing onto the base branch for syncing only should rarely be a problem. IMHO it's good practice to rebase the feature branch regularly to detect conflicts early. Many people use merges to sync with the base branch, but that usually just causes a lot of mess in the history of the feature branch.

> Overall, it is very tricky to split the commits up into nice chunks. I guess we could do a "only formating" commit and a "migration/fix" commit. But finer granulation will be very time consuming (and IMO without much benefit).

Yeah it's tricky. The best way to avoid these problems is to not do formatting changes and other changes at the same time. The better way would be to use auto-formatting for the web files and do a separate global format fixup PR.

Anyway, my main concern is to avoid having commits with `.`, etc. as description in the final git history. Apart from that, feel free to do as you like and if you have questions or need help with the git workflow, feel free to ask.

caco3:
Thanks for the tips, I will give it a try.
I think I will use this branch as a playground and once we are happy, adapt it in a new branch. Some non-related changes I already moved out.

PR #323, opened 2024-04-22
Enhance UI minification

caco3:
- Fix the minification of the JSON files.
- Embedded the lamda function directly into the sort function. This is needed for a successful minification.
- Simplified the debug host usage. It now gets automatically applied when the file is opened locally in a webbrowser.

This saves 4.7 kbytes of flash.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,7 @@ src/settings-override.h
 src/settings-custom-override.h
 node_modules/
 html/REST_API.yaml
+html/*.min.html
+html/locales/*.min.json
+html/js/*.min.min.js
+html/js/swaggerInitializer.min.js
--- a/html/management.html
+++ b/html/management.html
@@ -1351,15 +1351,15 @@
 		});
 	}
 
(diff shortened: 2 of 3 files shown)


tueddy:
The automatic installation of minify_html worked perfectly:
```
Trying to Install required module: minify_html
If this failes, please execute "pip install minify_html" manually.
C:\Users\user\.platformio\penv\Scripts\python.exe -m pip install minify_html
Collecting minify_html
  Downloading minify_html-0.15.0-cp311-none-win_amd64.whl.metadata (18 kB)
Downloading minify_html-0.15.0-cp311-none-win_amd64.whl (2.4 MB)
   ---------------------------------------- 2.4/2.4 MB 6.3 MB/s eta 0:00:00
Installing collected packages: minify_html
Successfully installed minify_html-0.15.0
GENERATING HTML HEADER FILES
...
```
I've checked the reduced size for JSON here, left side before minification:
grafik

Good catch!

PR #325, opened 2024-04-23
Fix js minification

caco3:


tueddy:
This PR fixes the regression, the info dialog works again:

grafik

PR #328, opened 2024-04-28
Web UI Enhancements

caco3:
- Migrated to Bootstrap 5.3.3
- Added dark mode. It gets persisted in the local storage of the webbrowser
- Modal dialogs do not get closed when clicking outside of the dialog
- Cover and control buttons are now centered
- The cover image size is limited to 400px
- Remove redundant texts (automatically replaced by i18n texts)
- Automatically scroll log window down to bottom on opening it or a reload
- Moved the header bar buttons into a hamburger menu
- The Control button bar and the sliders (with their volume buttons resp. icons now always stay on the same line (down to a screen width of 320 px) 
- On a page reload, the previously selected tab gets reloaded
- Added icons to the jstree context menu
- Fixed/added some missing translations
- Improved consistency of the overall web UI 
- Many more minor tweaks

Note: The firmware grows approx 1.3 kBytes due to this PR.

Diskussion: https://forum.espuino.de/t/darkmode/2893

### DIFF
--- a/html/locales/de.json
+++ b/html/locales/de.json
@@ -13,8 +13,8 @@
     "ok": "OK",
     "infos": "Infos",
     "darkmode": "Dunkelmodus",
-	"restartinfo":"ESPuino wird jetzt neu gestartet...\n\nDie Seite lädt sich nach dem Neustart automatisch neu.",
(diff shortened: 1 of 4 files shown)


tueddy:
Just tested it, it already looks very good!

There are still a few visual inconsistencies, but some of them were already there before, so they might be fixed later. 
The dark theme is primarily intended to reduce the brightness of Papa's mobile phone & this has already been done well.

Let us know if you think you are ready..

PR #329, opened 2024-05-02
Check if title only contains valid UTF8 and truncate it if needed

caco3:
See https://forum.espuino.de/t/eszett-ss-im-id3-tag-titel-macht-den-websocket-kaputt/2988


tueddy:
The encoding problem seems to be solved by updating the audio library. 
I'll close it here for now, feel free to reopen it if the problem occurs again.

PR #334, opened 2024-09-09
change default settings to match "ESPuino-mini 4Layer" board

tueddy:
https://forum.espuino.de/t/espuino-mini-4layer/1661


tueddy:
This requires more work or a separate HAL

PR #336, opened 2024-09-13
fix sleep after playlist not toggling nightmode

dneukirchen:
Wrong property was passed to the sleep handler.


tueddy:
@dneukirchen First thanks for your contribution!
Please make a PR against the DEV-branch next time.

PR #344, opened 2024-10-06
Slider range and html cleanup

Trainbird:
This replaces the 3rd party library bootstrap slider range and makes use of the bootstrap v5 slider.

It also includes the changes from [html-cleanup](https://github.com/Trainbird/ESPuino/tree/html-cleanup), as it was forked from there.


Trainbird:
Too many changes, going for smaller requests step by step

PR #346, opened 2024-10-14
Removal of non-psram HALs

biologist79:
As discussed here: https://forum.espuino.de/t/dev-branch/1835/642


tueddy:
+1 & good documentation update!
I will test it the next days.

PR #352, opened 2024-11-26
LPCD: Only wake up if card is present in NVS

fox34:
Do not wake up ESPuino if detected card is not present in NVS, [as discussed here](https://forum.espuino.de/t/lpcd-nur-starten-wenn-erkannte-karte-in-nvs-hinterlegt-ist/3345/11).

Note: as of now, this PR does not handle encrypted ISO-15693 cards (e.g. original Tonies).


fox34:
`System_Init_LPCD` could be renamed to `System_Init_Rfid_Namespace` for further clarification, opinions on this?

PR #355, opened 2024-12-11
use saved SSID as LAST_SSID, when no successfull connection was possible

nona00:
I observed when no successfull conecction was possible (e.g. when mac filter on fritz box is active) after reboot no connection was made. In the patch we use the saved SSID as LAST_SSID


tueddy:
Cherry picked in DEV branch, @nona00 thanks for contribution!

PR #360, opened 2025-04-15
make mqtt work again

Joe91:
just used static-casts to resolve the issues with arduiono 3 and mqtt.
Seems to work fine this way for me.


tueddy:
Looks good!
Unfortunately I don't have an MQTT broker here and can't test it. Can someone else do it?

biologist79:
Ich schaue mal, dass ich das morgen teste.

PR #367, opened 2025-05-15
Draft: Remove audio task

Joe91:
need some more testing, but already functional...


Joe91:
will create a clean PR for that...

PR #374, opened 2025-06-26
No more Audio-Task

Joe91:
This time without all the steps in between.
When we decided on how to handle the parallel upload this can be finished with one more commit...


Joe91:
I updated the SPI-LED-branch and merged it into this branch again.
LED-Task still gets paused during upload (increases upload significantly).

Upload-Speeds (full folder with 125 MB and 11 Files):
While Audio-Playback: 438.83 KB/s (with LED-Task-running 384.40 KB)
While Paused: 492.23 KB/s (with LED-Task-running 424.37 KB/s)

Audio-Lib was also updated to 3.4.0.
Async Webserver was also updated to the latest release.
Fix for RFID-Taks-Pause, since this could cause a Watchdog-Error the way it was paused before.
No more flickering of the LEDs, no more separated Audio-Task.

PR #390, opened 2026-02-10
Allow setting the minimum volume in the UI

lukas227:
Make the web interface a bit more silent (less "action completed successfully messages when there are just async data fetches") and add the option to set the minimum volume.


lukas227:
Hrn, sorry - had developed & tested this against master and only cherry-picked onto `dev` without testing. I'll close this for now and create a new one once I've properly tested things.

PR #392, opened 2026-02-28
Allow setting the minimum volume in the UI

lukas227:
Make the web interface a bit more silent (less "action completed successfully messages when there are just async data fetches") and add the option to set the minimum volume.

Originally https://github.com/biologist79/ESPuino/pull/390 but now properly updated against the dev branch and tested.

### DIFF
--- a/html/locales/de.json
+++ b/html/locales/de.json
@@ -349,6 +349,7 @@
 		"language": "Sprache",
 		"volume": {
 			"title": "Lautstärke",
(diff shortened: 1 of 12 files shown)


biologist79:
Sorry for the quite long delay.
Released it finally along with this commit: https://github.com/biologist79/ESPuino/commit/4110092de00924c3bf856307c9298ca1bade61cc.

Thanks for your contribution!

PR #394, opened 2026-03-05
Bug fix: Suppression of sound artifacts on the speaker during Bluetooth playback

joker-mik:
This solution works for me. I no longer get sound glitches when connected via Bluetooth. Switching to wired headphones that may be plugged in also works.
 https://forum.espuino.de/t/soundfetzen-aus-lautsprecher-bei-verbundenen-bluetoothkopfhoerern/4324/3


joker-mik:
There were too many changes in the pull request. I created a new one.
Fingers crossed
https://github.com/biologist79/ESPuino/pull/397

PR #395, opened 2026-03-06
Feature: ResumeOnSameRFID.

joker-mik:
If the ‘Don't accept same RFID twice’ option is enabled, this option causes playback to continue when the same RFID is placed on the reader again. 
I have also added a feature to the web interface that disables the ‘Don't accept same RFID twice’ option when ‘Pause when RFID tag is removed’ is enabled. I have also adjusted the explanatory text in this regard. According to the comment in the code, these options are mutually exclusive.

see also https://forum.espuino.de/t/neues-feature-wechsel-von-pause-auf-play-bei-akzeptiere-rfid-nur-einmal-resume-on-same-rfid/4341

### DIFF
--- a/html/locales/de.json
+++ b/html/locales/de.json
@@ -324,14 +324,14 @@
 		},
 		"sink": {
 			"title": "ESPuino als Bluetooth-Lautsprecher",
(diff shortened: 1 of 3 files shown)


joker-mik:
I marked it as a draft because I saw that clang completely changed the JSON language files. I'll undo that.

PR #396, opened 2026-03-12
audio_active was shadowed and actually never toggled after init

bebu:


biologist79:
Sorry for the long delay in getting back to you.
In general, please submit fixes against the dev branch, since changes are tested there before being merged into master.
In dev, the whole handling has since changed, so this PR is no longer applicable.

PR #402, opened 2026-04-14
Bug fix: Suppression of sound artifacts on the speaker during Bluetooth playback - Secound Edition

joker-mik:
The channel detection now takes place in Audioplayer.cpp.  The firmware build for TTGOT8 no longer fails either.

### DIFF
--- a/src/AudioPlayer.cpp
+++ b/src/AudioPlayer.cpp
@@ -62,6 +62,18 @@ static uint32_t AudioPlayer_HeadphoneLastDetectionTimestamp = 0u;
 static uint8_t AudioPlayer_MaxVolumeHeadphone = 11u; // Maximum volume that can be adjusted in headphone-mode (default; can be changed later via GUI)
 #endif
 
+static bool AudioPlayer_IsHeadphoneModeActive() {
+#ifdef HEADPHONE_ADJUST_ENABLE
+	const bool wiredHeadphoneConnected = !Audio_Detect_Mode_HP(Port_Read(HP_DETECT));
+#else
+	const bool wiredHeadphoneConnected = false;
+#endif
+
+	const bool bluetoothHeadphoneConnected = (System_GetOperationMode() == OPMODE_BLUETOOTH_SOURCE) && Bluetooth_Device_Connected();
+
+	return wiredHeadphoneConnected || bluetoothHeadphoneConnected;
+}
+
 // dummy class to allocate audio object in PSRAM if available
 class AudioCustom : public Audio {
 public:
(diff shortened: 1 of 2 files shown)


src/AudioPlayer.cpp
```diff
@@ -1596,5 +1611,11 @@ void audio_oggimage(File &file, std::vector<uint32_t> v) {
 
 // record audiodata or send via BT
 void audio_process_i2s(int16_t *outBuff, int32_t validSamples, bool *continueI2S) {
```
Joe91, line 1611: Hi, I don't understand your intention here. What's the benefit of adding this logic here, if the same logic is already in the Bluetooth_source_SendAudioData function

src/AudioPlayer.cpp
```diff
@@ -510,7 +522,9 @@ void AudioPlayer_SetupVolumeAndAmps(void) {
 	Port_Write(GPIO_HP_EN, true, true);
 	#endif
 #else
-	if (Audio_Detect_Mode_HP(Port_Read(HP_DETECT))) {
+	const bool headphoneModeActive = AudioPlayer_IsHeadphoneModeActive();
```
Joe91: if the boolean is used only once, why not call the fuction directly in the if()?

Joe91 (commented):
I don't understand the intention of the process i2s change. Rest is okay.

src/AudioPlayer.cpp
joker-mik: You're right, we can just as easily put the boolean in the if statement. I sometimes like to leave it out for debugging purposes. That's fine for me

src/AudioPlayer.cpp
joker-mik: I think Bluetooth_Source_SendAudioData(...) simply determines:
“Was I able to place this audio data into the Bluetooth ring buffer?”
audio_process_i2s(...) additionally determines:
“Should the same amount of audio data still be output locally via I2S afterwards?”

The previously logic was:

*continueI2S = !Bluetooth_Source_SendAudioData(outBuff, validSamples);

This means:
BT transmission successful  -> continueI2S = false
BT transmission failed -> continueI2S = true

This could be a problem.

If Bluetooth_Source_SendAudioData() returns false even briefly, e.g. because the ring buffer is full or a timeout occurs, then continueI2S = true.

Then these samples do continue to run locally via I2S. If the speaker amplifier is still active or cannot be switched off cleanly, you will hear precisely these brief artefacts.

I want to prevent the fallback logic from using I2S.

Perhaps that won't happen, and there are too many lines. I'm not sure about that

Joe91: Ah, now I understand it. Okay. Good idea. Your assumption is right.

src/AudioPlayer.cpp
joker-mik: Now, the if statement calls the function directly

joker-mik:
Corde was merged with the following pull request: https://github.com/biologist79/ESPuino/pull/407

PR #412, opened 2026-06-25
Feature/track seeking with rotary encoder

joker-mik:
### Add rotary encoder seek feature

Adds a seek mode controlled by the rotary encoder.  
When seek mode is triggered via the configured button action, encoder rotation adjusts the current playback position forward or backward.

The feature adds configurable seek behavior through the management interface:

- seek delay before playback resumes
- encoder turn threshold for seek steps

This improves navigation in long audio tracks such as audiobooks or radio plays without adding a separate playback command.

### DIFF
--- a/html/locales/de.json
+++ b/html/locales/de.json
@@ -87,7 +87,6 @@
 		},
 		"seek": {
 			"title": "Spulen (Drehimpulsgeber)",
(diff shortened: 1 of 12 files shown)


biologist79:
Implemented using https://github.com/biologist79/ESPuino/pull/436.

PR #413, opened 2026-07-06
RfidMfrc522: deterministic WUPA presence poll for pauseIfRfidRemoved

mgoodfellow:
I've rebased onto current `dev` (`d4cffb0`) and added a second commit — plus captured hard before/after evidence on a classic ESP32 (LOLIN D32 Pro) + RC522 clone. All timestamped serial traces below; each opens with the boot banner so the firmware revision is verifiable per run.

## A/B evidence, card sitting perfectly still for 10 minutes, `pauseIfRfidRemoved` on

| Firmware | `mfrc522Gain` | Spurious "RFID-tag removed" / 10 min |
|---|---|---|
| stock `dev` @ `d4cffb0` | 7 (default) | **20** — continuous pause/resume flapping for ~50 s, sub-second cycles |
| stock `dev` @ `d4cffb0` | 5 | 0 |
| this PR @ `8074410` | 7 (default) | **1** (single phantom at initial card contact — receiver saturation, see the tuning note in `docs/rfid-presence-detection.md`) |
| this PR @ `8074410` | 5 | **0** — plus 5/5 deliberate removal cycles all detected cleanly, pause on lift / resume on re-apply |

Two knock-on effects of the stock flapping worth highlighting, both in the traces:

1. **Flash wear** — every flap writes the play position to NVS (one
   `Write '#...' to NVS` per spurious removal).
2. **The box deep-sleeps itself mid-playback** — the pause/resume races desync
   the play/pause state, the inactivity timer arms while audio is audibly
   playing, and 10 minutes later: `Going to deepsleep due to inactivity-timer...`
   mid-track. On a kids' audiobook box this is "it turned itself off mid-story".

## Why the second commit (`Rfid: forget the last-seen tag when its assignment changes`) is part of this PR

The first commit fixes `lastValidcardId`'s lifetime (stock declares it inside
the task loop — uninitialised each iteration, so the "same card re-applied"
dedup effectively never fired). With the dedup *working*, a latent bug becomes
visible: re-assigning a tag in the web UI has no effect until reboot — the
re-applied card is short-circuited into play/pause and NVS is never re-read.
The existing guard (`Rfid_ResetOldRfid` on the assign path) is gated behind
`dontAcceptRfidTwice`, which `AudioPlayer.cpp` force-disables whenever
`pauseIfRfidRemoved` is on — i.e. unreachable in exactly the mode that needs it.

The second commit invalidates the reader's cache from every path that changes
what a tag means (WS + REST assign, mod-card, delete, NVS erase/restore, web-UI
play). Verified live in the same session, no reboot: card playing book A →
lifted → reassigned via `POST /rfid` to book B → re-applied → book B plays
immediately (`RFID-tag received` in the trace shows it reached the queue
instead of being deduped away).

Shipping the first commit without the second would hand every
`pauseIfRfidRemoved` user the reassign bug, so they travel together.

[trace-baseline-dev-d4cffb0.log](https://github.com/user-attachments/files/30085842/trace-baseline-dev-d4cffb0.log)
[trace-stock-dev-gain7.log](https://github.com/user-attachments/files/30085843/trace-stock-dev-gain7.log)
[trace-wupa-gain5.log](https://github.com/user-attachments/files/30085844/trace-wupa-gain5.log)
[trace-wupa-gain7.log](https://github.com/user-attachments/files/30085847/trace-wupa-gain7.log)

### DIFF
--- a/docs/rfid-presence-detection.md
+++ b/docs/rfid-presence-detection.md
@@ -1,0 +1,133 @@
+# RC522 constant-presence detection: root cause & fix
+
+Context: In `pauseIfRfidRemoved` mode, playback would flap (pause/resume) while a
+card sat perfectly still on the reader, and card-removal detection was unreliable
+on RC522 clones. During debugging we first tried papering over this with a growing
+"consecutive misses" debounce (3 → 8 cycles ≈ 400 ms). That treats the symptom.
+This note records the actual root cause and the fix.
+
+## Root cause: REQA never reliably re-detects a stationary card
+
+The old presence loop polled with `PICC_IsNewCardPresent()`. Under the hood that
+sends **REQA (`0x26`)**. From the MFRC522 library's own header:
+
+| Command | Opcode | Wakes cards in state… |
+|---------|--------|-----------------------|
+| `REQA`  | `0x26` | **IDLE only** ("Sleeping cards in state HALT are ignored") |
+| `WUPA`  | `0x52` | **IDLE and HALT** |
+| `HLTA`  | `0x50` | moves an ACTIVE card → HALT |
+
+The ISO-14443A state machine is: `IDLE → (REQA/WUPA) → READY → (Select) → ACTIVE → (HLTA) → HALT`.
+
+In `pauseIfRfidRemoved` mode the old code deliberately did **not** `HaltA()` after
+reading the card (`RfidMfrc522.cpp`, the `if (!gPlayProperties.pauseIfRfidRemoved)`
+guard around `PICC_HaltA()`). So the card was left in the **ACTIVE** state — and an
+ACTIVE card does not answer REQA. The loop then relied on a block of "voodoo"
+(from miguelbalboa/rfid issue #188) that re-ran `PICC_ReadCardSerial()`
+(anticollision + Select) on retries and checked for a magic `control == 13 || 14`.
+
+Whether that re-Select succeeds on any given poll depends on which state the card
+(and a sloppy clone's state machine) happens to be in at that instant. That is the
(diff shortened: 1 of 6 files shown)


mgoodfellow:
Just as a point, it turns out the default gain of 7 caused me all my initial problems, hence the creation of this to deal with it. Now its stable on most gains, but it appears that even the "stock" is pretty stable at 5 for close range (contact) cards. So maybe that's a better default for new users? 

Alternatively, adding a tooltip around the "pauseOnRfidRemoved" option to mention that a lower gain provides better stability for that mode, with the loss of tap range.

biologist79:
Mentioned this in my forum and asked for testing/opinions: https://forum.espuino.de/t/versuch-der-verbesserung-pause-when-rfid-removed-fuer-rc522/4609

The thing is: I don't use both RC522 and pausedIfRfidRemoved. So I cannot really assess, if your feature does things make any better. At all pausedIfRfidRemoved (which I dislike and would never implement again) is pretty complicated as every user builds up its own enclosure. So "environment" is pretty heterogeneous.

mgoodfellow:
That's fair. I built it this way for my kids so they can slot cards in themselves. They absolutely love it, but it had to be rock solid.

My enclosure inside looks like this:

image

You can see the RC522 mounted on the card slot with stand offs. You drop card in and works perfectly. I discovered this project because I was looking for an open source yoto/tonie box, as I disgree with their wasteful closed ecosystems.

mgoodfellow:
Fun fact, that's actually an ESP32-S3 build - the orange box has the Lolin D32 v2 pro build. They are effectively the same capabilities wise (other than loss of bluetooth on the S3 variant). Otherwise, I just cross compile to both :)

PR #414, opened 2026-07-06
LED: drive WS2812 via NeoPixelBus on ESP32-S3 and classic ESP32 (FastLED broken on Arduino 3.x / IDF 5.x)

mgoodfellow:
## What

Makes NeoPixel/WS2812 LEDs work on **both ESP32 variants** by driving the strip
with **NeoPixelBus** instead of FastLED: the **ESP32-S3** uses NeoPixelBus's
LCD/GDMA method, and the **classic ESP32** uses its I2S method (I2S peripheral
1). FastLED stays exactly as-is for the `CRGB`/`CHSV` math and all animation
code — only the driver (init/show/clear) is swapped. Animations are identical
on every target; only the output backend changes.

## Why

On Arduino core 3.x / ESP-IDF 5.x, **no FastLED WS2812 backend works on either
ESP32 variant.**

**ESP32-S3:**

| FastLED backend | Result on S3 / Arduino 3.3.8 / IDF 5.5 |
|---|---|
| clockless-SPI (what `Led.h` forces today) | hard-locked to the FSPI host → collides with SPI peripherals (SD/RFID) → `ESP_ERR_TIMEOUT` boot loop |
| RMT5 (FastLED 3.10.3) | `ESP_ERR_NO_MEM` in `led_strip_new_rmt_device` |
| RMT5 (FastLED 3.9.16) | strip created, then `ESP_ERR_INVALID_ARG` at `rmt_transmit` |
| legacy RMT4 (`-DFASTLED_RMT5=0`) | boot abort: `rmt(legacy): CONFLICT! driver_ng is not allowed to be used with the legacy driver` |

**Classic ESP32 (WROVER):**

| FastLED backend | Result on classic ESP32 / Arduino 3.3.8 / IDF 5.5 |
|---|---|
| SPI-DMA (`ClocklessSpiWs2812Controller`) | garbled output, then an `ESP_ERROR_CHECK` abort in `SpiStripWs2812::waitDone()` → boot loop, once SD + RFID already own both SPI hosts (HSPI/VSPI) |
| RMT | same legacy-RMT `driver_ng` conflict as the S3 row above |

NeoPixelBus's **RMT** method is unusable for the same reason as the legacy-RMT4
rows — through v2.8.4 it is built on the legacy RMT driver. So each target uses a
method that avoids RMT (and the SPI hosts):

- **S3:** LCD_CAM + GDMA (`NeoEsp32LcdX8Ws2812xMethod`) — no RMT, no SPI host.
- **Classic ESP32:** I2S (`NeoEsp32I2s1Ws2812xMethod`) on **I2S peripheral 1**,
  leaving I2S0 for the audio path — no RMT, no SPI host.

Both coexist with SD, RFID readers, and the audio I2S.

## Implementation notes

- The NeoPixelBus path in `Led.cpp` is now enabled for **all ESP32 targets**
  (was S3-only); the output method is chosen at compile time — LCD/GDMA on the
  S3, I2S1 on the classic ESP32. The S3-only IDF-5.5 `gpio_hal_iomux_func_sel`
  shim is guarded behind `CONFIG_IDF_TARGET_ESP32S3`.
- `Led.h` no longer forces `FASTLED_ESP32_USE_CLOCKLESS_SPI` on the S3 (the
  FastLED driver is no longer initialized on any ESP32, but the guard keeps the
  define correct for anyone experimenting).
- Small helpers `Led_InitStrip` / `Led_ShowStrip` / `Led_ClearStrip` in
  `Led.cpp` select NeoPixelBus (ESP32/S3) or FastLED (fallback) at compile time.
  Animations are untouched — they still fill the `CRGB` buffer.
- Global brightness and `TypicalSMD5050` colour correction are applied per-pixel
  in the show helper (NeoPixelBus has neither built in).
- IDF 5.5 removed `gpio_hal_iomux_func_sel()`, which NeoPixelBus ≤ 2.8.4 still
  calls in the LCD-method header; `Led.cpp` shims it around `PIN_FUNC_SELECT`
  (same operation).
- Fixes a latent bug on the way: the LED-count re-config path used to call
  `FastLED.addLeds` again without releasing the previous controller (leaked a
  controller/RMT channel per change). The new path deletes the strip first.
- `Led_LoadSettings` falls back to the compile-time LED counts when NVS holds a
  stored total of 0 (happens when settings were saved by a build with
  `NEOPIXEL_ENABLE` off) — a zero-pixel strip left the LEDs permanently dark
  with no error.
- Same class of bug for `dimmableStates`: an NVS value of 0 (also saved by a
  `NEOPIXEL_ENABLE`-off build) caused an `IntegerDivideByZero` panic in
  `Animation_Progress` at playback start; now guarded with the same
  fall-back-to-default pattern.

## Testing

- **ESP32-S3-DevKitC-1 clone (N16R8)**, 12-LED WS2812B ring on GPIO 21, RC522 on
  FSPI, SD (SPI mode) on HSPI: boot/idle/volume animations render correctly,
  RFID and SD unaffected, no boot loop.
- **Classic ESP32-WROVER (LOLIN D32)**, 12-LED WS2812B ring on GPIO 12, RC522 +
  SD both on SPI (both hosts occupied): FastLED boot-looped here; with
  NeoPixelBus/I2S1 it boots clean, logs `NeoPixelBus strip initialized
  (12 leds)`, renders correct colours and brightness, and audio/RFID/SD are
  unaffected.
- Compile-checked `esp32-s3-devkitc-1` and `lolin_d32_pro` with stock settings.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -36,6 +36,7 @@ lib_deps =
 	https://github.com/madhephaestus/ESP32Encoder.git#7501cb47ec702c43de3c64df56d0b10930b9bc70 ; v0.12.0
 	https://github.com/Joe91/ESP-FTP-Server-Lib.git#517a2c90876659bfd6692d1f464784a05296e8d8 ; v1.0.1
 	https://github.com/FastLED/FastLED.git#20667c3a6413ed46a828f78ec95fb57e58d753f8 ; v3.10.3
+	https://github.com/Makuna/NeoPixelBus.git#fd670ee534064ddf5765e4db63c213ba05bbe880 ; v2.8.4, WS2812 output on ESP32-S3 (see Led.cpp)
 	https://github.com/ESP32Async/ESPAsyncWebServer.git#d6aa7daf0823a41966b1841826cff015237d6e42 ; v3.11.1
 	https://github.com/bblanchon/ArduinoJson.git#77771d3c07668e01d8f52acb03910c1110bb373f ; v7.4.3
 	https://github.com/pschatzmann/arduino-audio-tools.git#b4b32dc849233e1c51424221951c9f5c576671e5 ; v1.2.4
--- a/platformio.ini
+++ b/platformio.ini
@@ -144,6 +144,10 @@ board_upload.flash_size = 8MB
 ;https://docs.platformio.org/en/latest/boards/espressif32/esp32-s3-devkitc-1.html
 board = esp32-s3-devkitc-1
 platform = espressif32
+; The ESP32-S3 has no Classic Bluetooth; ESP32-A2DP #errors on it, and the
+; LDF compiles the lib even with BLUETOOTH_ENABLE off (chain mode doesn't
+; evaluate #ifdefs). Bluetooth.cpp compiles to stubs without it.
+lib_ignore = ESP32-A2DP
 
 ; change microcontroller
 board_build.mcu = esp32s3
--- a/src/Led.cpp
+++ b/src/Led.cpp
@@ -18,27 +18,30 @@
 #include <atomic>
 #include <esp_task_wdt.h>
 
-#if defined(NEOPIXEL_ENABLE) && defined(CONFIG_IDF_TARGET_ESP32S3)
-	// On ESP32-S3 with Arduino 3.x / ESP-IDF 5.x none of FastLED's WS2812
-	// backends work (clockless-SPI claims the FSPI host, RMT5 fails in
-	// led_strip_new_rmt_device/rmt_transmit, legacy RMT4 conflicts with the
-	// driver_ng the core installs). FastLED is kept for the CRGB/CHSV math and
-	// the animation buffer only; the actual output goes through NeoPixelBus's
-	// S3 LCD/GDMA method. (NeoPixelBus's RMT method is also unusable here: as
-	// of 2.8.4 it is built on the legacy RMT driver, which aborts at boot with
-	// "CONFLICT! driver_ng is not allowed to be used with the legacy driver".)
+#ifdef NEOPIXEL_ENABLE
+	// FastLED's WS2812 backends don't work on this Arduino 3.x / ESP-IDF 5.x stack:
+	// on the S3 the clockless-SPI backend claims the FSPI host and RMT5 fails; on the
+	// classic ESP32 the SPI-DMA backend aborts (ESP_ERROR_CHECK in SpiStripWs2812::
+	// waitDone) once SD + RFID already own both SPI hosts. So the strip is driven
+	// through NeoPixelBus instead — FastLED is kept only for the CRGB/CHSV math and
+	// the animation buffer. NeoPixelBus's RMT method is unusable here too (2.8.4 uses
+	// the legacy RMT driver, which aborts under the core's driver_ng), so the output
+	// method is target-specific (see Led_InitStrip): S3 -> LCD/GDMA, classic ESP32 ->
+	// the I2S peripheral 1 method (I2S0 is left for the audio path).
 	#define LED_USE_NEOPIXELBUS 1
 
-	#include <esp_idf_version.h>
-	#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 5, 0)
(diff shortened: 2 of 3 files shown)


Joe91:
Can you maybe split this MR? there are some good checks and validations in the LED class, but for now I think we don't need the S3 support yet. 
Are you on a S3? Right now the main issue with the LEDs is and was performance and robustness. This was achieved by using the Clockless-Spi in DMA mode. The RMT was not robust enough for our Load.
There might be better and more performant solutions without using the FastLed-Lib (since it has quite some overhead). Would it be worth to investigate this further?

biologist79:
Neopixel in general is pretty critical for ESPuino. Don't know if I'm willing changing things for ESP32-WROVER that work.
There's no need (anymore) to test LOLIN D32 because as it lacks of PSRAM, ESPuino doesn't run safely on it.
ESP32-S3: How fast is SPI when it comes to webtransfer? Just asking because we're using SDMMC in order it's twice as fast as SPI and saves a GPIO.

mgoodfellow:
Hi, I'm using D32 v2 Pros, its the same ESP32 WROVER chips with 4MB PSRAM. I'm currently building with these:

https://www.wemos.cc/en/latest/d32/d32_pro.html
and
https://www.amazon.co.uk/dp/B0D69YJKNL?ref=ppx_yo2ov_dt_b_fed_asin_title

I have it running perfectly stable on both. The reason for this patch is that on the D32 boards, I wanted to use the onboard SD card slot which uses up an SPI lane and therefore blocks FastLED from booting. In the case of the S3 its a similar story.

Sorry for the flood of PRs, I used to do electronics many years ago, and I had a bunch of S3s left over from some home assistant work I was doing, and this escalated as a project for my kids. It was harder to source the WeMos D32 Lolin V2 boards from china, so I started the S3 port in the meantime. As it happens due to the parts I have lying around in my kit, I also needed this patch for my Lolin setup due to using SD card onboard (SPI) rather than external (MMC)

mgoodfellow:
Also just to say, it's such an awesome project! Thanks for creating it!

This is what I've spun up so far with the team (my kids)

image
image
image

I know you guys have your PCBs, but I really wanted to get it running on a variety of hardware I had lying around from other projects. My skills don't lie in C, hence Claude doing a lot of heavy lifting, but I do have a long background in SWE and engineering in general, so always up for a challenge and sharing of knowledge.

biologist79:
Have to admit I've just read "LOLIN D32" and didn't notice WROVER.
On "our" side there's more or less nobody that's using ESP32-S3. So currently it's not the project's focus to be honest.

mgoodfellow:
Yes, but this allows the WROVER dev board to be used with the onboard SD card slot, rather than an external SD MMC module being added. The onboard SD card slot on the Lolin D32 V2 Pro WROVER uses the SPI lane, so FastLED will fail at boot.

I understand if it isn't something that is interesting, but I just like the idea of open hardware. This just feels like such a brilliantly open bit of software, I've managed to use it (and deploy it) on both sets of hardware, in under a week.

I thought this would be compile time opt ins, so not affect any of your primary targets? Forgive me if I misunderstood and I can try to make it like that, but getting this NeoPixel LED ring was the hardest bit of the entire process for me getting it running on either of the dev kits I mentioned (classic ESP32 and the newer ESP32-S3).

I think it would be a shame if we didn't offer easy paths for new users to start up with a range of hardware, esp since (in the UK anyway), its a right pain to get hold of the WROVER variant as it seems to have been dropped in favour of the S3 variant.

biologist79:
Sorry, but I don't want to add code anymore, that I cannot test.
My main focus ist classic ESP32-WROVER with SDMMC. This works fine with fastled.

mgoodfellow:
I think it's a bit of a shame, considering the ESP32 WROVER is effectively EOL.

What also more interesting is the new ESP32-S31 variant being released shortly into full scale production, which has the classic BT as well as other new features. That would make a good home for this project.

This change is entirely gated off from your development on the ESP32-WROVER. You require a specific set of platform flags to enable it, which by default are not enabled. 

I actually need to use it for the ESP32-WROVER D32 Lolin V2 boards I have because I want to use the onboard SD card slot which conflicts with fastLed.

PR #415, opened 2026-07-06
WiFi setup/onboarding flow — timeout, i18n, failure diagnosis, live test

mgoodfellow:
## What

Four improvements to getting an ESPuino onto WiFi, from the low-level connect loop up to the setup page:

1. **Connect window 5 s → 15 s** so DHCP can finish on slow networks.
2. **Setup page honors the browser language** (+ a language selector) instead of always German.
3. **The box tells you *why* a connect failed** (wrong password / not found / rejected) instead of retrying silently.
4. **Credentials are tested live from the setup page** — no more save → reboot → hope.

### `3a4907e` — 15 s connect window
On VLAN'd / mesh networks (UniFi IoT SSIDs and similar) the DHCP lease regularly takes more than 5 s after association. The old 5 s window tore the connection down and moved to the next BSSID before the lease arrived, so the station associated repeatedly but never got an IP. Widened to 15 s.

### `ca0501e` — browser language + selector
The setup page already embedded full de/en/fr i18next locales, but initialized with no language detection and `fallbackLng: 'de'` — so every user got German regardless of browser settings. Mirrors `management.html`: resolve the language from `localStorage('language')` or the browser with English fallback, add a small selector to change it on the spot, and localize `document.title` (which sat outside the `localize('body')` scope).

### `bc2ad9d` — tell the user why a connect failed
The connect state machine polls `WiFi.status()`, which never surfaces the 802.11 disconnect reason — so a wrong password, an out-of-range network and a router-side rejection all looked identical: silent retries, then fallback to the setup AP with no explanation.
- **Wlan:** capture the disconnect reason via `WiFi.onEvent`, remember the SSID of the running attempt, persist both to NVS when a connect cycle fails (cleared again on the next successful connection). If no attempt was started because no saved network showed up in the scan, report `NO_AP_FOUND`.
- **Web:** new `GET /wifistatus` returning the coarse connection state plus the last failure (ssid, raw reason code, and a user-explainable class: `wrong_password` / `not_found` / `rejected` / `other`).
- **accesspoint page:** localized banner with that diagnosis on load, so after a failed attempt the user is told e.g. "the password is probably wrong" instead of guessing. Locale strings added for en/de/fr.

### `1d537d8` — test credentials live from the setup page
Replaces the save → reboot → hope onboarding flow. After saving, the setup page POSTs `/wifitest` and the box tries the credentials on the station interface **while the setup AP stays up** (`WIFI_AP_STA`). The page polls `/wifistatus`
and shows the outcome in place:
- **connecting:** progress note (survives the brief hiccup when the AP hops to the target network's channel)
- **failed:** the classified reason (wrong password / not found / rejected / associated-but-no-IP), same vocabulary as the last-failure banner; the AP stays up so the user can correct and retry immediately
- **success:** the obtained IP is shown for ~15 s, then the box closes the AP and continues as a normally connected station — no reboot, no vanishing portal, and the user knows where to find the box afterwards

The AP idle-close timer is suspended while a test is running, so the portal can no longer disappear under a slow typist mid-setup.

## Testing

- **Live on ESP32-S3 hardware (2026-07-06):** on an English-locale browser the setup page came up in English with the language selector present; a deliberately wrong password produced the classified failure banner; and a correct-credential entry connected successfully via the live test and reported the obtained IP. Verified on the actual box, not just in a browser.
- Platform-agnostic — touches only `Wlan.*`, `Web.cpp`, `html/accesspoint.html` and locales; no board- or S3-specific code. Compile-checked on `esp32-s3-devkitc-1`.

## Splitting (if preferred)

The commits are layered but separable. `3a4907e` (the 15 s timeout) is an independent, low-risk bugfix and is happy to go as its own quick-merge PR if you'd rather not take the setup-page feature work at the same time. `bc2ad9d`
and `1d537d8` are coupled (the live test reuses the failure classification and `/wifistatus`); `ca0501e` is standalone frontend.

### DIFF
--- a/html/accesspoint.html
+++ b/html/accesspoint.html
@@ -90,6 +90,7 @@
 			</select>
 		</div>
 		<h1 data-i18n="wifi.title"></h1>
+		<div id="connstatus" style="display:none;background:#f8d7da;color:#721c24;border-radius:4px;padding:10px;margin:10px 0;text-align:left"></div>
 		<div id="WiFiList"></div>
 		<label for="ssid" data-i18n="wifi.ssid.title"></label>: <br>
 		<input type="text" id="ssid" name="ssid" placeholder="SSID" required>
@@ -165,6 +166,7 @@
 			if (typeof localize === "function") {
 				localize('body');
 			}
+			renderWifiStatus();
 		});
 		document.querySelector('#langSel').addEventListener('change', (event) => {
 			const lang = event.target.value;
@@ -174,6 +176,26 @@
 			}
 		});
 
+		let wifiStatusData = null;
+		function renderWifiStatus() {
+			const box = document.getElementById("connstatus");
+			const failure = wifiStatusData && wifiStatusData.last_failure;
+			if (!failure || !i18next.isInitialized) {
+				box.style.display = "none";
+				return;
+			}
+			const cls = ["wrong_password", "not_found", "rejected"].includes(failure.class) ? failure.class : "other";
+			box.innerText = i18next.t("wifi.fail." + cls, { ssid: failure.ssid, code: failure.reason });
+			box.style.display = "block";
+		}
+		async function getWifiStatus() {
+			try {
+				wifiStatusData = await (await fetch("/wifistatus")).json();
+				renderWifiStatus();
+			} catch (error) {
+				console.log(error);
+			}
+		}
 		function selectWiFi(ssid) {
 			document.getElementById("ssid").value = ssid;
 			document.getElementById("pwd").focus();
@@ -245,6 +267,7 @@
 			getWiFiList();
 			updateWifiFields();
 			setupFullInterfaceLink();
+			getWifiStatus();
 		});
 
 		// Show the full management interface's URL and offer a "copy to clipboard" button. Uses the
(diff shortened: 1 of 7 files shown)


biologist79:
1) 15s: ok
2) AP/language: As you maybe noticed, Js/CSS stuff is now delivered by ESP32 directly. This being said: there's now more possibilities - even for access.html. Initially I had the idea to merge both access and management but because of the captive portal, which is somehow restricted, this ended up in crap. I described it here: https://forum.espuino.de/t/lokale-auslieferung-der-frameworks-statt-cdn-auslieferung/4580. Guess KI will translate it for you :-). However, just want to mention this as probably this could influence your solution.
3) Reject-reason: ok
4) Onboarding/credentials: sounds pretty interesting!

Would love to have this big one broken down into four smaller PRs.

mgoodfellow:
Awesome will take a look over the weekend - this had a flow where it shows you live connection details in the captive portal flow. Will assess your changes, the forum comments, and come back with another offering next week :)

mgoodfellow:
Closing this in favour of smaller PRs as requested

PR #429, opened 2026-07-14
accesspoint: fix the blank save button (locale key was renamed)

mgoodfellow:
The WiFi setup page's save button is localized via `data-i18n="submit"`, but the web-UI
locale restructure (731ac13) renamed that key to `wifi.saveButton`. Nothing on `dev`
provides `submit` any more, so the button renders with **no label at all** — the one
button a user has to press to get the box onto WiFi.

One line: point it at the key that exists.

Before:

pr0-before-page-r

After:

pr0-after-page-r

### DIFF
--- a/html/accesspoint.html
+++ b/html/accesspoint.html
@@ -123,7 +123,7 @@
 			<br>
 			<br>
 		</div>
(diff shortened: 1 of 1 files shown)


biologist79:
It already noticed this morning but didn't fix it yet.
Thank you :-)

PR #432, opened 2026-07-14
Wlan/Web/accesspoint: tell the user WHY a connection attempt failed

mgoodfellow:
The connect state machine polls `WiFi.status()`, which never surfaces the 802.11 disconnect
reason — so a wrong password, an out-of-range network and a router-side rejection all look
identical: silent retries, then a fallback to the setup AP with no explanation.

- **Wlan:** capture the disconnect reason via `WiFi.onEvent` and persist it (plus the SSID)
  to NVS when a connect cycle fails, cleared again on the next successful connection. Reasons
  raised by our *own* teardown (`ASSOC_LEAVE` / `AUTH_LEAVE` — the state machine calls
  `WiFi.disconnect()` before every retry) are ignored, otherwise they overwrite the reason the
  AP actually gave.
- **Web:** new `GET /wifistatus` returning the connection state plus the last failure (ssid,
  raw reason code, and a user-explainable class: `wrong_password` / `not_found` / `rejected` /
  `other`).
- **accesspoint:** a localized banner with that diagnosis on load, so after a failed attempt
  the user is told e.g. "the password is probably wrong" instead of guessing. Locale strings
  for en/de/fr.

Verified on hardware: a wrong password now yields `{"reason": 15, "class": "wrong_password"}`
and the banner says so.

pr3-failure-reason-wrong-password

### DIFF
--- a/html/accesspoint.html
+++ b/html/accesspoint.html
@@ -90,6 +90,7 @@
 			</select>
 		</div>
 		<h1 data-i18n="wifi.title"></h1>
+		<div id="connstatus" style="display:none;background:#f8d7da;color:#721c24;border-radius:4px;padding:10px;margin:10px 0;text-align:left"></div>
 		<div id="WiFiList"></div>
 		<label for="ssid" data-i18n="wifi.ssid.title"></label>: <br>
 		<input type="text" id="ssid" name="ssid" placeholder="SSID" required>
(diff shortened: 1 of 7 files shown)


biologist79:
Please review conflict(s).

mgoodfellow:
Conflicts resolved :)

PR #433, opened 2026-07-14
Wlan/Web/accesspoint: test WiFi credentials live from the setup page

mgoodfellow:
Replaces the save → reboot → hope onboarding flow. After saving, the setup page POSTs
`/wifitest` and the box tries the credentials on the station interface **while the setup AP
stays up** (`WIFI_AP_STA`). The page polls `/wifistatus` and shows the outcome in place:

- **connecting:** the form is disabled and the Connect button becomes a "Connecting …"
  indicator (survives the brief hiccup when the AP hops to the target network's channel)
- **failed:** the classified reason, same vocabulary as PR 3's banner. The AP stays up and the
  form is re-enabled, so the user can correct and retry immediately.
- **success:** the obtained IP is shown, then the box closes the AP and continues as a normally
  connected station — no reboot, no vanishing portal.

Two supporting fixes:
- The AP idle-close timer is suspended while a test runs, so the portal can't disappear under a
  slow typist mid-setup.
- On success the "just want the web interface?" card is repointed at the IP the box just
  obtained and relabelled. It advertises `http://192.168.4.1/management.html` (hardcoded,
  because the captive-portal DNS makes `location.host` unusable) — an address that dies the
  moment the AP closes, so the copy button was handing out a dead link.

Built on PR 3 — it reuses the failure classification and `/wifistatus`. Rebases down to its own
commits once that lands.

**Known intermittency:** in 7 correct-password runs on a weak link (~-74 dBm), one failed with
`AUTH_FAIL` and was reported as "password is probably wrong". The same credentials connect
instantly in plain STA, so it looks like a transient auth failure in `AP_STA` (the radio
time-slices between AP and station). It's recoverable — the AP stays up and a retry works —
but a retry-before-declaring-failure would harden it if you'd rather.

pr4-liveflow-wrongpw-2-connecting
pr4-liveflow-wrongpw-3-fail
pr4-liveflow-success-3-success

### DIFF
--- a/html/accesspoint.html
+++ b/html/accesspoint.html
@@ -91,7 +91,6 @@
 		</div>
 		<h1 data-i18n="wifi.title"></h1>
 		<div id="connstatus" style="display:none;background:#f8d7da;color:#721c24;border-radius:4px;padding:10px;margin:10px 0;text-align:left"></div>
-		<div id="teststatus" style="display:none;border-radius:4px;padding:10px;margin:10px 0;text-align:left"></div>
 		<div id="WiFiList"></div>
 		<label for="ssid" data-i18n="wifi.ssid.title"></label>: <br>
 		<input type="text" id="ssid" name="ssid" placeholder="SSID" required>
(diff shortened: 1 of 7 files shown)


biologist79:
a) Needs to be fixed as well as I merged the previous PR.
b) To be honest: I dislike this 15s thing from a PR a few hours ago (and should not have merged it). Just entered a wrong pwd for a given WiFi and it took like 1,5 mins until I was able to log into AP. Not acceptable, so we need to solve this immediately. Suggestion: Make this duration configurable at access and management (pointing to the same nvs enty) and it's 5 s as per default.

mgoodfellow:
Will rebase tomorrow, I thought that might happen with the stacked PRs :)

I think just revert the 15s merge. I made that change as the first patch I did when I first got my dev board to compile and i was having trouble getting it to connect. I don't think it was ever an issue, 5s should be more than plenty for DHCP, but it was modified and put in as a "fix" that stuck around because "it worked". I'm more acclimatised to this system now and I think it was a different issue I was having with UniFi masquerading as a DHCP issue.

Will open a revert PR for you first thing tomorrow.

biologist79:
You don't need to provide a revert PR. I already did it :-)
Rebase however would be fine.

PR #436, opened 2026-07-16
"hold a button + turn the encoder" gestures

mgoodfellow:
## What

Adds a modifier concept to the rotary encoder: **while a button is physically
held, turning the encoder runs that button's configured rotary action instead
of changing the volume**. One action per rotation direction, per button —
configured exactly like the existing short/long press assignments:
compile-time default in `settings.h`, runtime-overridable via NVS
(`btnRotCw0..5` / `btnRotCcw0..5`), the `/settings` API, and new dropdowns in
the web UI.

Shipped defaults (both opt-in, everything else `CMD_NOTHING`):

- **hold NEXT + turn** → seek forwards/backwards within the track
- **hold PLAY/PAUSE + turn** → LED brightness up/down (new commands, see below)

The use case driving this: a kids' audiobook box. Seeking inside a
45-minute chapter and tweaking LED brightness at bedtime were both
impossible from the device itself — the encoder only ever meant volume.

## How it works / collisions solved

All three were pre-existing behaviours that only become visible once a button
can be held for a reason other than its own action:

- **Short press fires on release** → letting go of a modifier would also
  toggle play/pause. **Long press fires at `intervalToLongPress` while still
  held** → holding NEXT to seek would fire `CMD_LASTTRACK` mid-gesture and
  jump to the end of the audiobook. Both are now swallowed — but *only if the
  encoder was actually turned* (`usedAsModifier`). A plain press/hold on the
  same button still behaves normally; the long action of a gesture-mapped
  button simply resolves on release (the same deferral `CMD_SLEEPMODE`
  already used).
- **Two-button combos** match on the latched `isPressed`, so a held modifier
  was a sitting duck for an accidental combo → combos involving a
  currently-held modifier are skipped.
- **`ESP32Encoder::getCount()` keeps an incomplete half-step across ticks**,
  so a gesture could inherit rotation from before the button went down (and
  the volume could inherit rotation from during the gesture) → the count is
  cleared whenever the held modifier changes.

New commands `CMD_BRIGHTNESS_UP`/`CMD_BRIGHTNESS_DOWN` (154/155). They clamp:
`Led_SetBrightness()` takes a `uint8_t` and does not bounds-check, so stepping
down from brightness 2 would otherwise wrap to ~255 — a dark bedroom goes to
full blast.

## Web UI

New "Hold button + turn encoder" table directly below Button assignments
(General → button/encoder tab): 6 rows × two command dropdowns (turn
left/right), reusing the existing command-select machinery. Only shown when
the firmware reports a rotary encoder. Locale strings for en/de/fr.

Two adjacent gaps closed on the way:

- 154/155 are selectable in all command dropdowns (they'd be firmware-only
  otherwise).
- The factory-defaults endpoint (`/settings?section=defaults`) now reports
  the compile-time rotary defaults via `Button_GetRotaryActionDefault()` —
  without this, "reset to factory settings" blanked the new selects and the
  next save persisted `CMD_NOTHING` for all twelve.

## Supporting fixes (first two commits — standalone, worth having regardless)

- **`AudioPlayer_SetVolume()` clamps instead of rejecting**: a fast rotary
  spin near the rails (e.g. 20 → 23 with max 21) discarded the whole change
  and did nothing.
- **Relative seeks accumulate** (`seekmode` was a single overwriteable enum
  consumed once per audio-loop iteration): N rapid seek commands used to
  collapse into one jump; now an accumulating atomic, so five detents = 5×
  the seek step.

## Testing

Hardware-verified on a Lolin D32 + RC522 running a 29-track audiobook:

- hold-NEXT seek back/forward mid-chapter, incl. rapid multi-detent seeks
- hold-PLAY/PAUSE brightness up/down, incl. clamping at both rails
- plain short/long presses on the same buttons unchanged; long press on a
  gesture-mapped button resolves on release; turning cancels it
- combo suppression while a modifier is held
- web UI: dropdowns populate from NVS, save round-trips (`rotCw3` 0 → 170 →
  0 verified over `/settings`), factory-defaults restore the compile-time
  values, en/de/fr all render
- `pio run` clean on `lolin_d32_pro`; clang-format 16 clean

Full page:
encoder-tab-full-en

DE:
rotary-gestures-de

EN:
rotary-gestures-en
 
FR:
rotary-gestures-fr

Popover:
rotary-gestures-en-popover

### DIFF
--- a/html/locales/de.json
+++ b/html/locales/de.json
@@ -97,7 +97,11 @@
 		"rotary": {
 			"title": "Drehimpulsgeber",
 			"reverse": "Drehrichtung umkehren",
-			"reverseExp": "Kehrt die Zähl-/Drehrichtung des Drehimpulsgebers um, z.B. falls Lauter/Leiser vertauscht ist."
+			"reverseExp": "Kehrt die Zähl-/Drehrichtung des Drehimpulsgebers um, z.B. falls Lauter/Leiser vertauscht ist.",
+			"gestures": "Taste halten + drehen",
+			"gesturesExp": "Solange eine dieser Tasten gehalten wird, löst Drehen am Drehimpulsgeber die zugewiesene Aktion aus, statt die Lautstärke zu ändern. 'Rechts drehen' ist die Richtung, die sonst lauter stellt. Die Lang-Aktion einer solchen Taste wird erst beim Loslassen ausgelöst - und gar nicht, wenn gedreht wurde.",
+			"ccw": "links drehen",
+			"cw": "rechts drehen"
 		}
 	},
 	"wifi": {
@@ -289,6 +293,8 @@
 					"151": "🌐 IP-Adresse ansagen",
 					"152": "🕒 Uhrzeit ansagen",
 					"153": "💡 Ambient Light umschalten",
+					"154": "🔆 LED-Helligkeit erhöhen",
+					"155": "🔅 LED-Helligkeit verringern",
 					"0": "🗑 Lösche Zuordnung",
 					"170": "⏯ Play/Pause",
 					"171": "⏮ Vorheriger Titel",
--- a/html/locales/en.json
+++ b/html/locales/en.json
@@ -97,7 +97,11 @@
 		"rotary": {
 			"title": "Rotary encoder",
 			"reverse": "Reverse direction of rotation",
-			"reverseExp": "Reverses the rotary encoder's counting/turning direction, e.g. if volume up/down is swapped."
+			"reverseExp": "Reverses the rotary encoder's counting/turning direction, e.g. if volume up/down is swapped.",
+			"gestures": "Hold button + turn encoder",
+			"gesturesExp": "While one of these buttons is held down, turning the encoder triggers the assigned action instead of changing the volume. 'Turn right' is the direction that raises the volume. The long-press action of such a button fires when it is released - and not at all if the encoder was turned.",
+			"ccw": "turn left",
+			"cw": "turn right"
 		}
 	},
 	"wifi": {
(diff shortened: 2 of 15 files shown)


biologist79:
"Seeking inside a 45-minute chapter and tweaking LED brightness at bedtime were both impossible from the device itself"

That's not entirely correct.
a) Seeking was already possible using button actions. But yes, it's slower.
b) LED brightness is dimmed automatically when modification card for sleep is applied.

However, will test it.

mgoodfellow:
Sorry, I do apologise for Claude's overly emotive language and statements at times! It can also state untruths as facts, I should have proof read the PR body more, my fault.

Yes, I know seek existed, it actually had a bug on fast presses which this fixes as well (as you can spin the rotary fast so it exposed it) :) I also saw there was another PR with the idea of rotary to seek which is what fed this idea, but with the idea of holding a modification button to trigger the different modes.

With the brightness, I know you have the concept of modification cards, the versions I'm building for my kids are more like the yoto boxes, where they have a library of rfid cards with stories, and they pick and choose and put them in to listen and remove to stop. But this means, i cannot tap a modification card while it is active. I have a similar patch I'm working on to extend the sleep functionality and chapter sleep mode  from button presses so that I can go into room and prep shutdown after X minutes. Also exposed over MQTT for integration with Home Assistant, but its in progress still, so it's not ready for PR or inclusion into the project.

Anyway, I digress, I just wanted something I could quickly hold a button and dim the lights for them while it was playing. Also, it opens the door with lots more configuration options, i thought it would be a cool addition :)

This is the coolest project ever, thanks so much for creating it 🥳

PR #438, opened 2026-07-17
AudioPlayer: opt-in periodic play-position checkpoint for long audiobook chapters

mgoodfellow:
### Problem
Mid-file position is persisted only on a **clean** transition: pause, card removal (`pauseIfRfidRemoved`), and clean shutdown (`savePosShutdown`). If power just disappears mid-file — dead battery, yanked plug, watchdog reboot — nothing is saved, and the newest NVS record is the **start of the file** (from the last track boundary). A long single-file chapter loses up to an hour.

### What
Opt-in periodic checkpoint. New general setting `SAVE_PLAYPOS_INTERVAL` (NVS `savePosIntv`, seconds, **default 0 = off**): every N seconds the current position is written to NVS, reusing the exact `AudioPlayer_NvsRfidWriteWrapper(...,
getAudioCurrentTime(), ...)` call the pause path already uses. 

Gated to keep NVS-flash wear opt-in:
- audiobook play-modes only (`saveLastPlayPosition`);
- current file **≥ 5 min** (short/split tracks already checkpoint per boundary);
- interval-gated + deduped (skip if position hasn't advanced).

Interval clock resets at track start (first write lands one interval in). Web UI number field under General → Options, live-applied (no reboot), en/de/fr, with the same cached-page guard as `sleepTracks`.

### Tested
Compile (stock HAL 4) + clang-format-16 clean, locales parse. HW (Lolin D32): at 20 s interval the NVS position advanced in clean 20 s steps over a 59-min chapter (`12 → 32 → 53 → 73 → 97`, live via `GET /rfid`); end-to-end power-cut resume
verified with the resume-seek fix.

NOTE: This builds on the resume fix in https://github.com/biologist79/ESPuino/pull/437

Happy to discuss placement or logic - maybe it's not wanted? I'm not sure. The idea for this came about because I wasn't aware it was actually resume that was broken for CBR MP3s. However, I have been thinking about it, and I feel like this still has value for non-battery users (like me!), so decided to polish it anyway:

image


biologist79:
Sounds reasonable. However, you should mention that lowering the interval leads to wear of flash. Maybe Claude knows how many writes ESP32's flash can "resist". Just to give a number to the users.

mgoodfellow:
Interesting info actually!

1. ESPuino runs a deliberately enlarged NVS partition — 0x40000 = 256 KB (64 × 4 KB sectors), vs. the 24 KB default. The CSV even calls it out. NVS does wear-levelling across all of it, so checkpoints are spread over the partition, not hammered at one address.
2. NVS is log-structured — a putString appends an entry; a flash erase only happens when a page fills and garbage-collects. So one checkpoint is a fraction of an erase cycle, not one erase.
3. The write is gated — audiobook-mode + file ≥ 5 min + your interval + a dedup that skips if the position hasn't advanced. Short/split-track playlists never checkpoint at all.

> The flash chips used on ESP32 modules (Winbond W25Q / GigaDevice GD25Q class SPI-NOR) are rated ~100,000 erase/program cycles per sector (Espressif quotes the same figure). That's the number to give users.

To put a floor on it: even pretending every single checkpoint were a full erase cycle at the same address (it isn't),  writes at a 60 s interval is ~69 days of *continuous, non-stop* playback before reaching the rated cycle count — and the real figure is far higher. At a few hours a day that's well beyond the life of the device.

Just adding the note to the settings page now, then will undraft this PR :)

PR #442, opened 2026-07-21
Fix: Run timer and expander ISRs in IRAM on Arduino-ESP32 v3

oberbichler:
Both onTimer (100 Hz timer) and PORT_ExpanderISR (GPIO interrupt attached via attachInterrupt) were compiled without IRAM_ATTR on Arduino-ESP32 v3 via a version-conditional. If either ISR fires while the flash cache is disabled (e.g. during an NVS/flash write) it would execute from flash and crash; PORT_ExpanderISR is the higher risk since it can fire at any time. Place both unconditionally in IRAM and simplify onTimer's forward declaration to match.

Port_AllowReadFromPortExpander is written in PORT_ExpanderISR and read in the main task, so mark it volatile; align the (unused) extern in Button.cpp with the new type.

### DIFF
--- a/src/Button.cpp
+++ b/src/Button.cpp
@@ -52,17 +52,13 @@ uint8_t gShutdownButton = 99; // Helper used for Neopixel: stores button-number 
 uint16_t gLongPressTime = 0;
 
 #ifdef PORT_EXPANDER_ENABLE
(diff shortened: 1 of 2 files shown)


biologist79:
Merged!
Thanks for your contribution.

PR #445, opened 2026-07-24
Add configurable default action for unknown RFID tags

tommybotten:
An unassigned tag currently only shows an error. Add an optional "Default for unknown cards" that plays a preconfigured item instead, wired end to end:

- RfidCommon: on an unknown tag, if "unknownPath" (NVS) is set and the device is in normal mode, auto-detect the path and play it - a file as SINGLE_TRACK, a directory as ALL_TRACKS_OF_DIR_SORTED - otherwise fall through to the existing error. Mirrors a normal music card: no Rfid_ResetLastTag, so the resting tag keeps being deduped instead of re-triggering the default in a loop.
- RfidUnknownDefault.h: the enable/mode decision as a pure, host-testable helper.
- Web.cpp: settingsToJSON read-back and the save handler (empty = disabled; the empty string is written outside the success chain since putString returns 0 for it).
- New "Default for unknown cards" text field in the general options card, plus de/en/fr locales.

Empty by default, so behaviour is unchanged unless configured. Only acts in normal mode, so an unknown tag still just escapes Bluetooth mode.

Translations by mistral.ai


biologist79:
Sorry, but this doesn't make sense to me.

PR #448, opened 2026-07-29
RfidMfrc522: don't treat a corrupted WUPA reply as a removed card

mgoodfellow:
The pauseIfRfidRemoved presence poll counted every non-OK WUPA result toward the removal debounce. But the MFRC522 distinguishes two very different outcomes: STATUS_TIMEOUT means nothing answered (the field is empty), while a transmission error -- parity/protocol/CRC out of ErrorReg -- means a card *did* reply and the frame was mangled. The second is evidence the card is still present, so counting it as a miss makes any reader with RF noise nearby pause every few seconds.

Only clean timeouts now count toward removal, with a 1.5s no-clean-answer backstop so a card lifted during a noise burst is still noticed.

Second fix, same failure: after declaring removal the card was left in HALT. Re-detection uses PICC_IsNewCardPresent() -> REQA, which HALTed cards ignore, so a spurious removal made the reader permanently blind to a card that never left -- playback stayed paused until it was physically lifted and re-applied. The field is now cycled after a removal so any card still on the antenna returns to IDLE.

### DIFF
--- a/src/RfidMfrc522.cpp
+++ b/src/RfidMfrc522.cpp
@@ -83,16 +83,19 @@ void RfidMfrc522_TaskReset(void) {
 
 // Deterministic "is a card still on the antenna?" poll used by pauseIfRfidRemoved
 // mode. The card is kept parked in the ISO-14443 HALT state between polls; WUPA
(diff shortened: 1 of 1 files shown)


biologist79:

PR #449, opened 2026-07-29
Web-UI: report settings that block saving instead of failing silently

mgoodfellow:
The general settings form spans several tab-panes, and a browser will not report a validation error on a control it cannot focus -- which includes anything in a hidden pane. Native validation therefore blocked the submit silently: Save did nothing, no message appeared, and the only clue was "An invalid form control with name='...' is not focusable" in the browser console. A single out-of-range value on a tab the user was not looking at made the whole settings page unsaveable with no way to find out why.

The form now carries novalidate and is checked in JS on submit: the offending settings are named in an error toast, and the tab holding the first one is opened (main tab and sub-tab both) so it can be seen, focused and corrected.

This affected me after upgrading firmware, and having some values not set.

image

### DIFF
--- a/html/locales/de.json
+++ b/html/locales/de.json
@@ -19,7 +19,8 @@
 		"success": "Aktion erfolgreich ausgeführt.",
 		"dropout": "Langsamer Stream, Aussetzer möglich.",
 		"saved": "Einstellungen gespeichert.",
-		"ledCountRestart": "Die Anzahl der LEDs wurde geändert. Der ESPuino startet neu, um das zu übernehmen."
+		"ledCountRestart": "Die Anzahl der LEDs wurde geändert. Der ESPuino startet neu, um das zu übernehmen.",
+		"settingsInvalid": "Speichern nicht möglich: Bitte zuerst diese Einstellungen korrigieren ({{fields}})."
 	},
 	"status": {
 		"connected": "Verbindung zum ESPuino: OK",
--- a/html/locales/en.json
+++ b/html/locales/en.json
@@ -19,7 +19,8 @@
 		"success": "Action performed successfully.",
 		"dropout": "slow stream, dropouts are possible.",
 		"saved": "Settings saved.",
-		"ledCountRestart": "The number of LEDs was changed. ESPuino will restart to apply it."
+		"ledCountRestart": "The number of LEDs was changed. ESPuino will restart to apply it.",
+		"settingsInvalid": "Cannot save: please correct these settings first ({{fields}})."
 	},
 	"status": {
 		"connected": "Connection to ESPuino: OK",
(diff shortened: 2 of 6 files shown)


joker-mik:
Thanks for tackling this — the general approach makes sense, especially for the case where an invalid setting on another tab silently prevents saving.

There are two edge cases ChatGPT think we should consider:

1. `form.querySelectorAll(':invalid')` can also return elements such as `<fieldset>`, not only the actual invalid form control. According to the [HTML spec for `:invalid`](https://html.spec.whatwg.org/multipage/semantics-other.html#selector-invalid), a fieldset also matches if one of its descendant controls is invalid.

So with:

```js
const invalid = Array.from(form.querySelectorAll(':invalid'));
```

`invalid[0]` may be the fieldset rather than the actual `<input>`, which means:

```js
invalid[0].focus();
invalid[0].reportValidity();
```

may target the wrong element.

It might be more robust to only collect actual controls that participate in constraint validation:

```js
const invalid = Array.from(form.elements)
    .filter(el => el.willValidate && !el.validity.valid);
```

`willValidate` is specifically intended to indicate whether a control participates in form validation:
https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#dom-cva-willvalidate

2. I think we should distinguish between **a field hidden because it is on another tab** and **a field hidden because its whole feature/section is currently inactive**.

For a field on another tab, the current behavior is exactly what we want:

```text
invalid + hidden tab
→ open tab → focus field → show validation error
```

But for something inside a conditionally hidden section, e.g. battery / critical-voltage settings, the user may have no way to make the field visible and therefore no way to correct it.

In that case I think the control should probably not participate in constraint validation while the section is inactive:

```text
inactive section
→ don't validate its controls until the section becomes active
```

One option would be to disable those controls while the section is inactive. Disabled controls are explicitly excluded from constraint validation by the HTML spec:
https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#barred-from-constraint-validation

Alternatively, `required` / other constraints could be toggled together with the visibility state. Which option is better probably depends on whether disabled values still need to be included when the settings are serialized.

So overall I like the direction of this PR. I would just make sure we handle these two cases separately:

* hidden **tab** → reveal it and focus the actual invalid control
* inactive/hidden **feature section** → don't validate controls the user cannot currently edit

A small browser test for both cases would probably make this quite robust.

mgoodfellow:
Hi @joker-mik - thanks for the feedback, made some updates and added some tests! 

Updated this PR to address both validation edge cases:
- Invalid fields are now collected from form.elements using willValidate && !validity.valid. This ensures we target the actual form control rather than an invalid ancestor such as a <fieldset>.
- Conditionally unavailable sections are disabled while inactive, so their hidden required fields do not block saving.
- This applies to the optional NeoPixel and battery sections, plus the critical-voltage subsection when automatic shutdown is disabled.
- Activating a section now enables its fieldset at the same time it becomes visible.
- Existing hidden-tab behavior remains unchanged: the relevant main/sub-tab is opened, then the invalid control is focused and its native validation message shown.
- Added a headless Chrome regression test covering both hidden-tab navigation and inactive-section validation.

PR #454, opened 2026-08-13
Apply max volume settings without restart

joker-mik:
This change applies the configured speaker and headphone maximum volume immediately after saving the settings.

### Why?

Yesterday I asked my little son to turn the music down.

That worked great — for about two minutes. 😄
Shortly afterwards, he was back at maximum volume.

So I decided to show him that I had the upper hand and reduced the maximum volume through the web interface.

Well... nothing happened.

The setting was saved correctly, but the new maximum volume only became effective after the next restart. At that point, my supposedly clever parental intervention wasn't quite as impressive anymore.

And that's why this PR exists. 🙂

### What changed?

Previously, changes to `maxVolumeSp` and `maxVolumeHp` were stored in NVS but only became active after a restart.

With this change, the runtime maximum volume is updated immediately after the settings have been successfully saved.

If the current volume exceeds the newly configured maximum, it is reduced immediately.

The existing speaker/headphone mode handling remains intact, and the change avoids re-running the amplifier setup.

### DIFF
--- a/src/AudioPlayer.cpp
+++ b/src/AudioPlayer.cpp
@@ -621,6 +621,22 @@ void AudioPlayer_SetMaxVolumeSpeaker(uint8_t value) {
 	AudioPlayer_MaxVolumeSpeaker = value;
 }
 
+void AudioPlayer_ApplyMaxVolumes(uint8_t speaker, uint8_t headphone) {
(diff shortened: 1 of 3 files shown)


biologist79:

PR #455, opened 2026-08-13
Led: use FastLED FLEX_IO for ESP32 and ESP32-S3

mgoodfellow:
## Summary

Replace the forced FastLED clockless-SPI LED backend with FastLED's explicit
FLEX_IO channel API.

The selected hardware backend depends on the ESP32 target:

- Classic ESP32 uses I2S1 with DMA.
- ESP32-S3 uses LCD_CAM with GDMA.
- Other targets retain the existing `FastLED.addLeds()` path.

FastLED is pinned to a specific post-3.10.5 revision containing the required
FLEX_IO support.

## Why

ESPuino currently forces `FASTLED_ESP32_USE_CLOCKLESS_SPI` for NeoPixel output.

That backend competes for an SPI peripheral. On configurations where the SPI
host is already used by the SD card, initializing or refreshing the LED strip
can time out and cause boot failures or resets.

Falling back to RMT is not a reliable solution on the current Arduino 3.x /
ESP-IDF 5.x stack:

- The legacy RMT driver conflicts with the new IDF RMT driver.
- The RMT5 path can fail during channel allocation or transmission setup,
  particularly on ESP32-S3.

FastLED's FLEX_IO API allows us to select peripherals intended for parallel
output without consuming the SD card's SPI host:

- I2S1 on classic ESP32
- LCD_CAM/GDMA on ESP32-S3

This also keeps LED output separate from ESPuino's normal audio path, which
uses I2S0.

## Approach

`Led_InitStrip()` now creates a FastLED clockless channel explicitly:

- `Bus::FLEX_IO`, instance 0
- WS2812 800 kHz timing
- Existing color order, correction, dithering, brightness, and animation
  behavior remain unchanged

The target-specific FastLED bus traits are enabled only for:

- `CONFIG_IDF_TARGET_ESP32`
- `CONFIG_IDF_TARGET_ESP32S3`

Unsupported targets continue using the existing `FastLED.addLeds()` call.

A compile-time assertion currently limits the FLEX_IO path to the configured
WS2812B chipset. This fails clearly rather than silently emitting incorrect
timing if a custom build selects another chipset.

## Classic ESP32 FastLED correction

The pinned FastLED revision's classic ESP32 I2S implementation required a
small transport correction found during hardware testing.

The PlatformIO pre-build hook applies an exact, fail-closed patch that:

- feeds native 16-bit parallel DMA samples
- configures `tx_bits_mod = 16`
- uses the 16-bit single-channel FIFO mode
- routes lanes through `DATA_OUT8..23`
- adds reset-low data before and after the WS2812 frame
- stops transmission at DMA EOF to prevent stale FIFO data from being replayed

The hook is idempotent and tied to the pinned FastLED source. If that source
changes and the expected code no longer matches, the build fails instead of
silently applying a partial patch.

## ESP32-S3 task stack

FastLED's LCD_CAM initialization uses more stack than the previous LED backend.
The existing 3072-byte LED task stack overflowed during the first
`FastLED.show()`, causing a boot loop before Wi-Fi startup.

The ESP32-S3 LED task therefore uses a 6144-byte stack. Classic ESP32 retains
the existing 3072-byte allocation.

## Compatibility

For standard classic ESP32/WROVER ESPuino configurations:

- LED output uses I2S1.
- Audio continues using I2S0.
- SPI SD continues using HSPI.
- SD_MMC configurations are independent of both.

Therefore this should work for WROVER users whether they use SPI SD, SD_MMC,
or no SD SPI bus at all.

A custom configuration that already reserves I2S1 for another peripheral may
conflict and would need separate consideration.

## Validation

Hardware:

- Classic ESP32/D32 with a 12-pixel WS2812 ring
  - verified correct colors and independent pixel output
  - verified the previous duplicated/corrupted pixel behavior is fixed
- ESP32-S3 N16R8 with a WS2812 ring on GPIO 21
  - verified LCD_CAM/GDMA backend initialization
  - verified stable startup without stack overflow
  - verified SD, Wi-Fi, HTTP, audio, and RFID startup

Builds:

- `pio run -e lolin_d32_pro`
- `pio run -e lolin_d32_pro_sdmmc_pe`
- local ESP32-S3 DevKitC N16R8 environment

The resulting maps confirm:

- classic ESP32 links `ChannelEngineI2sEsp32Dev`
- ESP32-S3 links `ChannelDriverLcdClockless`
- neither target links the previous NeoPixelBus experiment

### DIFF
--- a/patchFastLedI2s.py
+++ b/patchFastLedI2s.py
@@ -1,0 +1,194 @@
+from pathlib import Path
+
+globals()["Import"]("env")
+env = globals()["env"]
+
+
+def replace_exact(path, old, new):
+    text = path.read_text(encoding="utf-8")
+    if new in text:
+        return
+    if text.count(old) != 1:
+        raise RuntimeError(
+            f"FastLED I2S patch no longer matches {path}; update the pinned patch before building"
+        )
+    path.write_text(text.replace(old, new), encoding="utf-8")
+
+
+fastled = Path(env.subst("$PROJECT_LIBDEPS_DIR")) / env.subst("$PIOENV") / "FastLED"
+engine = fastled / "src/platforms/esp/32/drivers/i2s/channel_engine_i2s_esp32dev.cpp.hpp"
+peripheral = fastled / "src/platforms/esp/32/drivers/i2s/i2s_peripheral_esp32dev_esp.cpp.hpp"
+
+if not engine.is_file() or not peripheral.is_file():
+    raise RuntimeError(f"Pinned FastLED dependency is missing from {fastled}")
+
+replace_exact(
+    engine,
+    "namespace fl {\n\n",
+    "namespace fl {\n\n"
+    "// Keep the one-shot I2S FIFO fed with low samples throughout the WS2812\n"
+    "// reset interval.\n"
+    "constexpr size_t kWs2812ResetTailBytes =\n"
+    "    (TIMING_WS2812_800KHZ::RESET * 6400000ULL / 1000000ULL) * sizeof(u16);\n\n",
+)
+
+replace_exact(
+    engine,
+    """    // Size the scratch buffer: max per-lane byte count across all
+    // in-flight channels. Three regions (see packScratchBuffer):
+    //   [0, 2W)        — 32-bit DMA samples (4 bytes per pulse), the
+    //                    region actually handed to transmit()
+    //   [2W, 3W)       — raw wave8 pulse pairs (2 bytes per pulse)
+    //   [3W, 3W + 16b) — 16-lane-strided transpose input
+    // where W = wave8I2s1EncodedFrameSize(bytes_per_lane).
+""",
+    """    // Size the scratch buffer for the 16-bit parallel DMA stream plus reset
+    // preamble/tail and the lane-strided transpose input.
+""",
+)
+
+replace_exact(
+    engine,
+    """    const size_t wave8_size = wave8I2s1EncodedFrameSize(bytes_per_lane);
+    const size_t output_size = wave8I2s1Encoded32FrameSize(bytes_per_lane);
+    const size_t input_size = 16 * bytes_per_lane;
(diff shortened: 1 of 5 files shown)


mgoodfellow:
The FLEX_IO branch is conceptually cleaner:

- One CRGB buffer.
- FastLED handles correction, brightness, dithering, and transmission.
- No per-frame conversion into a second library.
- It uses the direction FastLED itself appears to be moving toward.

However, those benefits are outweighed by the current dependency situation:

- It pins an unreleased FastLED commit.
- Classic ESP32 output was broken in that commit.
- We need patchFastLedI2s.py to rewrite FastLED internals during every build.
- A future FastLED change intentionally breaks the build until the patch is revisited.
- The pinned FastLED version produced a roughly 263 KB larger S3 image and used about 2.2 KB more internal RAM than the NeoPixelBus baseline.
- S3 required an additional task-stack adjustment.
- It currently supports only the explicitly mapped WS2812B timing.

The NeoPixelBus implementation is less elegant because it keeps FastLED for animation calculations and copies the result into NeoPixelBus for transport. It also manually reproduces brightness, correction, and gamma behavior. But it is a more self-contained and established workaround using released library code.

WROVER Compatibility

Both approaches use I2S1 on classic ESP32. Whether the SD card uses SPI, SD_MMC, or no SPI lane does not materially change the comparison.

They should both work for normal WROVER configurations where:

- audio uses I2S0;
- I2S1 is otherwise unused;
- LEDs are WS2812-compatible.

This is the result of comparing the 2 different approaches (with the closed PR here: https://github.com/biologist79/ESPuino/pull/414

However, I will keep an eye on the upstream FastLED to see what they do with it.

I don't think locking down the software to only work on one ESP32 variant is the right approach. I think flexibility of platforms and different hardware configurations would be great for a OSS project like this. Especially considering the new release of the ESP32-S31 variants with more ram, and classic bluetooth making a return.

PR #461, opened 2026-09-14
Improve Bluetooth status, scanning, reconnect handling and web UI

joker-mik:
## Summary

This PR improves Bluetooth Source handling and the Bluetooth section of the web interface.

The main goals are to make Bluetooth configuration easier to understand, provide better feedback about the current connection state, improve device discovery and connection handling, and make reconnects more reliable.

It also fixes an issue where Bluetooth GAP discovery could continue running after an A2DP Source connection had already been restored. In that situation discovery traffic continued in parallel with audio playback and could interfere with the newly established stream.

### Bluetooth web interface

- Reworked the Bluetooth configuration page to make the available operating modes easier to understand and select.
- Added a dedicated mode selector for:
  - Normal mode
  - Bluetooth Source / headphones
  - Bluetooth Sink / speaker
- Added a visible connection status indicator for Bluetooth Source mode.
- Improved the layout of the Bluetooth settings and scan controls.
- Improved the device list for smaller screens and mobile devices.
- Added and updated translations for German, English and French.

### Connection status

- Added a backend endpoint that exposes the current Bluetooth connection state.
- The web interface uses this information to show whether the Bluetooth Source is currently connected.
- Status polling is limited to Bluetooth Source mode while the Bluetooth tab is actually open, avoiding unnecessary requests while the user is on other pages.

### Device scanning

- Improved Bluetooth device discovery and scan-state handling.
- Scan results now contain the information needed to select and connect to a discovered device directly from the web interface.
- Improved handling of scan callbacks and restoration of the A2DP library callback after a manual scan.
- Improved handling of devices that are already connected or known.

### Connection and retry handling

- Reworked connection retry handling so retries represent actual connection attempts.
- Prevented duplicate disconnect events from consuming additional retry attempts.
- Improved state cleanup after successful connections and failed connection attempts.
- Improved switching from one Bluetooth device to another.
- Intentional disconnects while switching devices no longer consume retry attempts for the new target.
- Connection targets and pending connection state are tracked more explicitly.

### Reconnect / GAP discovery fix

The A2DP library may start its own Bluetooth GAP inquiry while trying to automatically reconnect to a previously paired device.

Previously, this inquiry was not represented by the web UI's `scanInProgress` state. As a result, calling the normal scan cleanup code after a successful reconnect did not necessarily stop the library-owned inquiry.

This could leave Bluetooth discovery running for several seconds after the A2DP connection had already been restored and audio playback had started.

The connection handler now explicitly cancels GAP discovery after a successful Bluetooth Source connection:

- Manual web scans still use the existing scan cleanup path so their callback and UI state are restored correctly.
- Library-owned discovery during auto-reconnect is explicitly cancelled once the connection succeeds.
- `ESP_ERR_INVALID_STATE` is ignored when no discovery is active.

This prevents discovery traffic from continuing alongside the newly established A2DP audio stream.

### API and documentation

- Extended the REST API documentation for the Bluetooth status functionality.
- Updated the related web UI translations.

## Reconnect test

The reconnect behavior was tested with Bluetooth headphones by:

1. connecting the headphones,
2. switching them off,
3. waiting for the disconnect,
4. switching them back on,
5. allowing ESPuino to reconnect automatically.

Before the fix, log output showed repeated `Device found` messages continuing even after the Bluetooth connection had already reached `Connected` and audio playback had started.

After the fix, GAP discovery is stopped when the connection succeeds and no longer continues in parallel with playback.

The reconnect scenario now behaves correctly in the tested setup.

## Compatibility

The changes are limited to Bluetooth handling, the Bluetooth-related web interface, translations and the corresponding REST endpoint/documentation.

The branch has also been updated to the current `dev` branch before opening this PR.


tbaumann:
Tested. UI and workflow work great and are a huge improvement.

However audio was choppy in bt headphones mode. I will re-flash dev to compare if that is due to changes in this branch.

PS: Complete, build from source

tbaumann:
> However audio was choppy in bt headphones mode. I will re-flash dev to compare if that is due to changes in this branch.

Can't reproduce it either way. Probably because I had my phone connected to the headphones too. 
-> Probably a fluke

tbaumann:
[0e467ff](https://github.com/biologist79/ESPuino/pull/461/commits/0e467ffe84b13759143e3153e09a63589c31493f) A rebase would be better IMO.

tbaumann:
+1

joker-mik:
Thanks for testing. And you are right a rebase is the better way

